PR #38, opened 2022-09-22
Chroma / ZMQ

MorganAskins:
- Needs GDML geometry export


MorganAskins:
Closing this PR as it might not be the best way to both include GDML support and the zero mq socket transportation but the branch itself can be used to serve as an example for a future attempt (pending changes to how chroma reads gdml)

PR #89, opened 2023-08-17
Incorporate sophisticated quenching routine v2

masmiley:
Reviving #72 which got closed when my repo got deleted in the move to the rat-pac org.


MorganAskins:
Can we stuff the `quenching` folder into `physics`, and the `quadrature` folder into ... somewhere ... `util`?

masmiley:
Okay, I think these latest commits take care of the outstanding items for this PR, so it should be good to go in.

In working on these changes, it did raise the question of whether parameter setting by DB or a Messenger should take precedent (and whether Messenger-set parameters should be overwriting the DB too), but I think that's going to need to be handled in a separate PR since it's a broader question.

PR #90, opened 2023-08-17
More improvements to logging v2

masmiley:
Reviving #82 which got closed when my repo got deleted in the move to the rat-pac org.


MorganAskins:
Any way we can fix the conflict so that this can be merged? I changed the branching we're pointing at to be the next-update instead of main.

masmiley:
Should be fixed now

PR #112, opened 2023-11-19
fix decay chain issue reading in branching ratio and several other quantities

tannerbk:
Doubles should be read in using %lf according to the man page for [fscanf](https://cplusplus.com/reference/cstdio/fscanf/). I have no idea why this only broke now.


JamesJieranShen:
Hey @tannerbk , I have been working on #114 and we seem to have made similar changes to these files. Just wanted to let you know since merging both may result in conflicts...

hsong7273:
LGTM!

tannerbk:
Closed in favor of #114.

Issue #97, opened 2023-09-20
Float comparison should be flagged when compiling


masmiley:
There are several places where exact equality is used for floats in RAT-PAC, which can have machine-dependent effects. Warnings should be generated when compiling for any of these instances.

tannerbk:
Should be fixed by #114 can you confirm?

PR #114, opened 2023-11-20
Enable (some) compile-time warnings, change `stlplus` header location

JamesJieranShen:
I looked into this while investigating Issue #97 reported by Max. I noticed that currently, all compiler warnings are suppressed in  [CMakeList.txt](https://github.com/rat-pac/ratpac-two/blob/c768e1abe331ac2593dd25358e352aa20471c10b/CMakeLists.txt#L27), which could prevent many potential bugs to slip through the gaps when the compiler could have caught them. Based on the commit history, it appears that the warning messages were originally turned off to suppress excessive warnings from `stlplus`, a piece of external source code included in the RAT build chain. This PR seeks to mitigate this problem by doing the following:
1. Global compile warnings are turned ON. I am only using the default rule sets (with additional warning flags) for now. Ideally we should move to even stricter rulesets like `-Werror -Wall -Wextra`, as well as the warning options proposed by Max in the original issue, but these options produced so many warnings that i have decided to not include them at the moment.
2. Errors related to stlplus source files are suppressed in the stlplus and RAT_DICT build targets directly.
3. stlplus headers are now moved to their own directory. Previously, both RAT and stlplus headers were located at `include/RAT`, and were simultaneously included during build. stlplus headers are now moved to `include/stlplus`, and this folder is now included during build as system headers such that warnings coming from them are suppressed. 
4. Fixed a couple of warnings emitted by the compiler after they are turned back on. Particularly in `daq/NoiseProc`, `gen/DecayChain` and `gen/BetaFunction`
5. Modified the stlplus source files to fix a couple exception warnings that won't go away even when warnings are turned off.

I have included one commit from PR #111 to ensure that these two requests do not conflict from each other, since they both make edits to the stlplus source files.


tannerbk:
I reverted this because I realizes it breaks the compilation of EosSimulations without corresponding changes there. And worth updating the example RatpacExperiment as well.

PR #125, opened 2024-03-29
Add PMT channel to PMTInfo

tannerbk:
PMT ID populates in the order they're built in the geometry. Channel number can be used to make the actual channel number in the detector, to map from electronics channel to detector position.

### DIFF
--- a/src/ds/include/RAT/DS/PMTInfo.hh
+++ b/src/ds/include/RAT/DS/PMTInfo.hh
@@ -41,7 +41,7 @@ class PMTInfo : public TObject {
   }
 
   virtual void AddPMT(const TVector3& _pos, const TVector3& _dir, const int _type) {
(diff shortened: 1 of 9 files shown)


src/ds/include/RAT/DS/PMTInfo.hh
```diff
@@ -40,7 +41,7 @@ class PMTInfo : public TObject {
   }
 
   virtual void AddPMT(const TVector3& _pos, const TVector3& _dir, const int _type) {
-    AddPMT(_pos, _dir, _type, "", 1.0, 0.0, 0.0);
+    AddPMT(_pos, _dir, _type, -1.0, "", 1.0, 0.0, 0.0);
```
JamesJieranShen, line 44: type definition is an `int`, here it is initialized as a float.

PR #131, opened 2024-04-30
Add seperate branches in ntuple for MCPE frontend time and hit time

JamesJieranShen:
**Potentially breaking!**

Old `mcPETime` in ntuple is the time that front-end electronics see a detected photon, which includes timing effects of the PMT. This PR modifies this behavior such that the branch `mcPETime` now records the time that a photon hits the PMT. The front-end time is written to a new branch named `mcPEFrontEndTime`.

Since this change assigns an old variable name to a new definition, previously-implemented analysis may need to be modified to account for it.

### DIFF
--- a/src/io/include/RAT/OutNtupleProc.hh
+++ b/src/io/include/RAT/OutNtupleProc.hh
@@ -96,7 +96,7 @@ class OutNtupleProc : public Processor {
   std::vector<int> mcpmtid;
   std::vector<int> mcpmtnpe;
   // MCPE
-  std::vector<double> mcpetime;
+  std::vector<double> mcpehittime;
   std::vector<double> mcpefrontendtime;
   std::vector<int> mcpeprocess;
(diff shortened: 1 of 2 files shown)


tannerbk:
We can announce this on the next call and merge then.

gmwendel:
Can you explain why you renamed the variable "mcPETime" to record mc photon absorption times, rather than creating a new variable like "mcPEIncidentTime"? I'm concerned about potential confusion for users accessing these ntuple files, especially regarding which version of "mcPETime" is used during the transition to the new ratpac version and when referring to older file versions.

JamesJieranShen:
per comment by @gmwendel I've changed the name of the old branch to `mcPEHitTime`, in accordance to the naming of the variable internal to RAT::DS::Photon. I personally have no strong opinion to this either way. If others do, I am happy to revert this change.

JamesJieranShen:
@tannerbk Can we merge this?

PR #132, opened 2024-05-06
Added average timing response of R11780 and R7081 PMTs to PMTTRANSIT.ratdb

hbjamin:


JamesJieranShen:
Thanks @hbjamin ! Couple of things that is worth checking:
1. Should this go to EosSimulation or rat-pac2? Might be a question for @tannerbk 
2. I believe the PMTTRANSIT table must directly correlate to indices in the PMT table. So entry `r11780` will not be used in simulation as-is.
3. We should think about what to do with the original Hamamatsu  measurements for the 7081. Either remove them or create a seperate PMT model for them.

PR #134, opened 2024-05-22
Memory Leak in the PMTWaveformGenerator code.

JamesJieranShen:
demote a pointer, and place the class on the stack directly.


MarcFBergevin (commented):
Standard change from pointer to non-pointer.

PR #139, opened 2024-06-17
Add nested tube array geometry (optical fibers)

WilfS:
Optical fibers (such as wavelength shifting fibers) are not explicitly defined in ratpac, despite their prevalence in physics experiments. One could create such a fiber in ratpac by constructing several concentric tubes with relevant materials, but when simulating experiments with thousands of fibers this can lead to undesirably large geometry files.

To address this issue, this update will add the nested tube array geometry, that allows construction of an array of three concentric tubes (a core, an inner tube and an outer tube). The length, position, and direction of each nested tube can be set individually. Materials and radii are set for the whole array.

Validation has been performed to ensure the nested tube arrays are correctly placed in the geometry and have the same optical properties as concentric tubes manually placed in the geometry. A validation macro and geometry file have been added that demonstrate how the nested tube arrays are placed.

The rat DS has been updated to include nested tube array info such as direction, length and material. This is similar to the PMTInfo class.

No extra materials have been added that would be useful when using nested tube arrays as optical fibers. We hope to add this later when we have more accurate properties.

The documentation has been updated to include the nested tube array geometry.

### DIFF
--- a/doc/users_guide/processes.rst
+++ b/doc/users_guide/processes.rst
@@ -128,7 +128,7 @@ below.
 Wavelength Shifting
 ```````````````````
 There are a few ways of doing bulk wavelength shifting in RAT. The default
-behavior is for GLG4Scint to handle opticalphotons as well as charged
+behavior is for GLG4Scint to handle optical photons as well as charged
 particles. Alternatively, you can also let GLG4Scint handle the primary
 scintillation, then use Geant4's G4OpWLS process or the custom BNLOpWLSModel
 to do the reemission.
--- a/doc/users_guide/data_structure.rst
+++ b/doc/users_guide/data_structure.rst
@@ -13,7 +13,7 @@ The Ratpac data structure is defined in the `src/ds` directory and lists
 everything under the `RAT::DS` namespace. An instance of the data structure is
 defined in the `RAT::DS::Root` object. The data structure is tree-like, with each
 instance of `RAT::DS::Root` containing a list of `RAT::DS::MC` objects which contain
-the Monte Carl truth infomration, and a list of `RAT::DS::EV` objects which contain
+the Monte Carl truth information, and a list of `RAT::DS::EV` objects which contain
 the reconstructed event information (usually after going through a processor that
 simulates the detector response).
 
--- a/doc/users_guide/database.rst
+++ b/doc/users_guide/database.rst
@@ -63,7 +63,7 @@ Normally, tables and fields are all you have to think about, but RATDB also
 addresses an additional complication: overriding constants.  It is a common use
 case to have default values of constants, values which are only valid in
 certain time intervals and can change (like the optical properties of the
-scintilator), and user-specified values which are intended to override
+scintillator), and user-specified values which are intended to override
 everything.
 
 RATDB handles this by internally grouping tables into three ''planes''.  The
@@ -79,7 +79,7 @@ When an item is requested, RATDB will attempt to locate it in the user plane
 first, then the time plane, and finally the default plane.  Note that this is
 all handled in the background.  You simply request the index_of_refraction
 field in the MEDIA[acrylic] table, and RATDB figures out the appropriate plane
-from which to retreive the data.
+from which to retrieve the data.
 
 How do I load data into RATDB?
 ``````````````````````````````
--- a/doc/users_guide/geometry.rst
+++ b/doc/users_guide/geometry.rst
@@ -80,6 +80,7 @@ Allowed types:
  * tubearray - Array of tubes
  * lgarray - Array of tubes where one end has the PMT face cut out 
  * pmtarray - Array of PMTs
+ * nestedtubearray - Array of three nested tubes. Useful to simulate optical fibers
  * waterboxarray - Array of standard cubitainer water boxes
  * extpolyarray - Array of extruded polygonal solids
  * bubble - Collection of bubbles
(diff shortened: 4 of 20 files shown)


doc/users_guide/processes.rst
JamesJieranShen, line 1: This should not change -- `opticalphoton` is the name of the particle class in G4.

JamesJieranShen (commented):
This looks mostly ok with me on the geom construction perspective. I will wait on comments from others before merging.

JamesJieranShen:
Argh.. please fix the formatting as well.

doc/users_guide/processes.rst
WilfS: Apologies. I've now changed this back. Thanks for checking.

WilfS:
Just pushed changes that are clang-formatted. Sorry, I missed that requirement.

WilfS:
I've just added some commits relating to adding the ability to make the core of the nested tubes a sensitive detector. This is particularly useful when using the nested tubes as optical fibers. The hit information from the sensitive detector gets saved to the rat DS every event and can be accessed by processors.

MarcFBergevin:
I have successfully compiled the PR with no issues. However when I run the following macro:

`rat --vis macros/vis_nestedTubeArrays.mac -o ~/nested -l ~/nested.l`

I obtain the following error:

`GeoNestedTubeArrayFactory: Constructing volume fibers
DB: Field GEO[fibers].outer_material lookup failure. Does not exist or has wrong type.`

MarcFBergevin:
I have switched values from `material_outer` to `outer_material` in the geo file and I was able to get an output (also for inner and core). I am not too certain I understand the following output:

Screenshot 2024-09-13 at 8 58 34 AM

Is this the same output on your end?

WilfS:
Apologies i forgot to update the macro.

Yes that output is expected. It's demonstrating the construction of fibres. The different colors are for the different construction methods. One with the old method of defining individual tubes and the other with the new array method.

MarcFBergevin:
Thank you, that makes sense!

I did a quick test to test out the features with for four fibers. The visual output looks very nice. However, I am having difficulties accessing time and hit location in the outroot file (The error might be on my side). Have you cross-checked that hits are recorded for a test case scenario?

WilfS:
The fibers are sensitive detectors. Any hits that occur in them are saved as a hit collection in the ratds. I haven't added them to the outroot file, but they're easily accessed in processors so could be added

MarcFBergevin:
Apologies, I should have made a clearer statement. I did notice that the ntuple root files were not filled out and I was going to propose to you to add that in (potentially as an optional output to the ntuple root files). I was referring to the `/rat/proc outroot` (instead of `/rat/proc outntuple` files) which records the ratds in a root format.

By modifying the MC you essentially have already filled out the outroot root files (or ds files). However when I query via the TBrowser I get empty values:
TBrowser

~71000 photon bombs event where generated in the scintillator material with 4 fibers for my test. I can forward the macro to you. The setup looked like this.

fiberSetup

I expected to see some hits, but could not confirm any hits in the ratdb root files.

MarcFBergevin:
I have tried to add the sensitive detector option:

`{
name: "GEO",
index: "fibers",
enable: 1,
valid_begin: [0, 0],
valid_end: [0, 0],
mother: "outer_tank",
type: "nestedtubearray",
core_r: 0.47,
inner_r: 0.485,
outer_r: 0.5,
pos_table: "cable_pos",
orientation: "manual",
outer_material: "aluminum",
inner_material: "glass",
core_material: "mirror",
sensitive_detector: "/mydet/fibers",
#drawstyle: "solid",
color: [0.8,0.0,0.0,0.8]
}`

However, I am still not seeing any registered hits at this point.

JamesJieranShen:
> We've changed the example macro to use some example WLS materials that we've added to ratdb. With these materials photons will actually get absorbed in the core of the nested tubes and be recorded by the sensitive detectors.
> 
> Looking at the outroot file, I see entries for the `GetMCNestedTubeHitCount()` method, so the hits are getting saved to the ratpac DS. Yet, if I look at the hit information using `GetHitTime`, I get errors saying
> 
> ```
> Error in <TTreeFormula::DefinedVariable>: Unknown method:photon.GetHitTime() in vector<RAT::DS::MCNestedTubeHit> 
> Error in <TTreeFormula::Compile>:  Bad numerical expression : "ds.mc.nt.photon.GetHitTime()"
> ```
> 
> and if I try using the `hitTime` variable it says
> 
> ```
> Error in <TTreeFormula::DefinedVariable>: hitTime is not a datamember of vector<RAT::DS::MCNestedTubeHit>
> Error in <TTreeFormula::Compile>:  Bad numerical expression : "ds.mc.nt.photon.hitTime"
> ```
> 
> I may just have things setup wrong, as I get the same errors when I try to retrieve hit information from the MCPMT class.

Hey Wilf, you might need vectored versions of your data structures added to `LinkDef.hh`. Something like `#pragma link C++ class vector < RAT::DS::MCNestedTubeHit>;`. I think this will let cint (therefore ROOT) know about how to deal with vectors of the new DS objects.

WilfS:
Hi James. I believe I'm already doing that. And if I had forgotten to do it it wouldn't explain why the same thing is happening for the MCPhoton info. I must be doing something wrong such that cint isn't getting the info from linkdef

WilfS:
I've just added the requested feature from Marc where the nested tube hits are written to the ntuple output if requested.

If `/rat/procset include_nestedtubehits` is included in the macro, then nested tube info will be saved to the meta TTree and nested tube hit info will be saved in the output TTree.

tannerbk:
LGTM. @WilfS @gmwendel is this ready to go? I left one small comment about some variables that I don't think are used.

src/io/include/RAT/OutNtupleProc.hh
```diff
@@ -150,6 +166,9 @@ class OutNtupleProc : public Processor {
   std::vector<int> hitPMTID;
   std::vector<double> hitPMTTime;
   std::vector<double> hitPMTCharge;
+  std::vector<double> hitPMTDigitizedTime;
```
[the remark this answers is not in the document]
WilfS: Good spot. I think some of the variables are old ones that somehow didn't get removed when I merged in the latest changes from the old branch. I'll go in and clean them up.

PR #142, opened 2024-07-03
Fix Materials.cc when scaling scint pdf

### LINKED ISSUE #141: Spectrum of properties as a function of wavelength
When defining the emission spectrum (or re-emission in the case of WLS materials) as a function of the wavelength, Ratpac uses the spectrum defined in the RATDB file, but it converts the wavelength to energy. This approach does not accurately reproduce the property, which is defined as a function of the wavelength in the database.
To correctly define the property as a function of wavelength, which is then used as a function of energy, it is necessary to normalise the spectrum by the square of the wavelength. This requirement can make the definition of this type of property confusing and prone to errors.
Ideally, Ratpac should handle the conversion of the spectrum from being a function of wavelength to a function of energy.

[PropertiesWavelength.pdf](https://github.com/user-attachments/files/16032230/PropertiesWavelength.pdf)

gmwendel:
Fixes #141.  This only applies the Jacobian element to optical ratdb entries with the 'wavelength' key option specified (wavelength_opt == 1).  The 'dy_dwavelength' keyword (wavelength_opt == 2) will need a closer look as I am unfamiliar with what y stands for in this context.
[EOS_ratpac_jacobian.pdf](https://github.com/user-attachments/files/16088445/EOS_ratpac_jacobian.pdf)

### DIFF
--- a/src/geo/src/Materials.cc
+++ b/src/geo/src/Materials.cc
@@ -460,6 +460,7 @@ G4MaterialPropertyVector *Materials::LoadProperty(DBLinkPtr table, std::string n
       if (E_value != 0.0) {
         double lam = E_value;
         E_value = CLHEP::twopi * CLHEP::hbarc / (lam * CLHEP::nanometer);
+        if (wavelength_opt == 1) p_value *= 1.0 / (E_value * E_value);
         if (wavelength_opt == 2) p_value *= lam / E_value;
       } else {
         warn << "Materials property std::vector zero wavelength!" << newline;


JamesJieranShen:
As discussed in Slack, the correct approach here is indeed to use the `dy_dwavelength` mode. We make distinctions since not all properties (as function of energy) requires re-weighting when converted to energy. Example being RINDEX, ABSLENGTH, etc. Close for now. Feel free to re-open if features beyond what is provided in the current options are required. Thanks!

gmwendel:
Yes, after closer inspection this is not the correct way to handle this.  Fixing this quirk will take a bit more effort as I have now mentioned in the parent issue.

PR #143, opened 2024-07-15
Companion changes for the Eos InHDF5Producer

JamesJieranShen:
- Allow producers to be added in experiments built on top of ratpac2, such as EosSimulation.
- Fixes segfaults caused by OutNTuple when there are zero MC particles.
- Refactored and extended WaveformAnalysis:
  - Allow it to take in different ratdb entries by specifying their indices.
  - To allow for both Constant fraction discriminator and simple threshold timing
  - Use the simple threshold timing to allow simple step-like traces coming from digitized triggers to be analyzed.

### DIFF
--- a/src/daq/src/WaveformAnalysis.cc
+++ b/src/daq/src/WaveformAnalysis.cc
@@ -4,8 +4,10 @@
 
 namespace RAT {
 
-WaveformAnalysis::WaveformAnalysis() {
-  fDigit = DB::Get()->GetLink("DIGITIZER_ANALYSIS");
+WaveformAnalysis::WaveformAnalysis() { WaveformAnalysis(""); }
+
+WaveformAnalysis::WaveformAnalysis(std::string analyzer_name) {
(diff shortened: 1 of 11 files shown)


src/daq/src/WaveformAnalysis.cc
```diff
@@ -46,6 +51,33 @@ void WaveformAnalysis::RunAnalysis(DS::DigitPMT *digitpmt, int pmtID, Digitizer
...
+  fVoltageRes = (fDigitizer->fVhigh - fDigitizer->fVlow) / (pow(2, fDigitizer->fNBits));
+  fTimeStep = 1.0 / fDigitizer->fSamplingRate;  // in ns
+
+  fDigitWfm = fDigitizer->fDigitWaveForm[pmtID];
+
+  CalculatePedestal();
+  double trigger_threshold = fThreshold;
+  double trigger_lookback = fLookback;
+  try {
+    trigger_threshold = fDigit->GetD("trigger_voltage_thrershold");
```
tannerbk, line 64: Typo in naming.

tannerbk (commented):
Looks good to me.

PR #150, opened 2024-07-23
Add channel number to the DS for the PMT objects

tannerbk:
This allows one to grab the channel number directly rather than relying on using the meta data to create a map from PMT ID to channel number.


JamesJieranShen:
I'm not sure if it is necessary to add all the pmt_channel fields in MCPMT/PMT objects. The information is available in RUN::PMTInfo. I do agree though that we should write this to the ntuple. This can be done with a PMTInfo lookup.

JamesJieranShen:
Merge conflict means this may need to be rebased?

tannerbk:
Yeah I'm going to close this for now and simplify according to your suggestions.

PR #160, opened 2024-08-15
Simulated dead channels

JamesJieranShen:
- Add new channel status class that incorporates any hardware channel information. Analogous to the ChanHWStatus class in other rat experiments. 
  - By default, apply timing offset of 0 and turn PMT online, so backward compatibility should be ensured.
- Disable photoelectron generation in Gsim if a channel is not set to `online` .
- Minor refactoring to CMake compilation flags to suppress stlplus-induced warnings.

### DIFF
--- a/ratdb/RUN.ratdb
+++ b/ratdb/RUN.ratdb
@@ -2,6 +2,8 @@
 {
 name: "RUN",
 run_range: [0, 0],
+// index into channel_status table to use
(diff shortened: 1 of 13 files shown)


JamesJieranShen:
Another commit to add the relevant info into ntuples. I am only writing out information that correspond to channels with PMTs plugged in, not any utility channels.

PR #162, opened 2024-08-22
Moving towards supporting multiple waveform analysis algorithms

JamesJieranShen:
Motivation: We would like to eventually support multiple types of waveform analysis to be performed on digitization traces. To accomplish this, we need to have the following

- [x] Convert the current WaveformAnalysis class to a RAT::Processor. This allows user to swap between different analysis algorithms, similar to how users can swap between existing DAQ modules. 
- [ ] However, there's a lot of features that should ideally be shared between all the waveform analyzers. Examples being: calculating the pedestal and computing the number of threshold crossings. These features currently exist in the old WaveformAnalysis class, and should perhaps be re-factored into static methods that can be called by any processor that requires it.
- [x] There could be information about the digitization that come upstream of the waveform analysis. For example, one might want to set the local trigger time via the producer, and reconstruct the waveforms themselves later (as we do in Eos). So let's use a std::map keyed by PMTID, instead of std::vector to keep track of the pmts. This allows the `GetDigitPMT` method to create an empty digitPMT if needed, and retrieve the existing PMT if it has already been instantiated. This also removes a potentially confusing indexing of digitPMTs in `DS::EV`, since the index position of the digitPMTs in the vector may not agree with the value written in `digitPMT.id`. 
- [ ] DigitPMT should support multi-pe scenarios. Perhaps using the same structure as MCPMT, where you can have multiple MCPEs associated with a single DigitPMT

### DIFF
--- a/ratdb/DIGITIZER_ANALYSIS.ratdb
+++ b/ratdb/DIGITIZER_ANALYSIS.ratdb
@@ -26,4 +26,5 @@ fit_window_high: 15.0,
 // Fit shape
 lognormal_shape: 0.15,
 lognormal_scale: 10.5,
+apply_cable_offset: 0,
 }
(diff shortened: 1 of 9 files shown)


JamesJieranShen:
I think the discussion I've posted here should continue to be worked on, but this PR is self-contained and can be merged now to facilitate further work downstream. Changes made to Waveformanalysis is non-breaking, and all old behaviors should still work with this class.

PR #163, opened 2024-08-24
Fix untriggered EV info in ntuple

llebanowski:
Assign invalid values to the two EV branches currently not 'reset' for untriggered events.  
Without this change, untriggered events inherit `nhits` values from the previous triggered event.

### DIFF
--- a/src/io/src/OutNtupleProc.cc
+++ b/src/io/src/OutNtupleProc.cc
@@ -467,7 +467,10 @@ Processor::Result OutNtupleProc::DSEvent(DS::Root *ds) {
     outputTree->Fill();
   }
   if (options.untriggered && ds->GetEVCount() == 0) {
+    // EV information
     evid = -1;
+    subev = -1;
+    nhits = -1;
     triggerTime = 0;
     timeSinceLastTrigger_us = 0;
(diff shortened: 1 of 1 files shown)


src/io/src/OutNtupleProc.cc
```diff
@@ -467,7 +467,10 @@ Processor::Result OutNtupleProc::DSEvent(DS::Root *ds) {
     outputTree->Fill();
   }
   if (options.untriggered && ds->GetEVCount() == 0) {
+    // EV information
     evid = -1;
+    subev = -1;
+    nhits = -1;
```
JamesJieranShen, line 473: Thanks for catching this! Do you think nhits should be set to 0 instead of -1?

llebanowski: i think -1 is safest as an invalid value for this int object.  
though seemingly unlikely, one could imagine a case where there are actually 0 hits in a detector while a trigger is still issued for some reason.

JamesJieranShen: ok, fine. I guess downstream can also override that behavior if needed. LGTM!

PR #164, opened 2024-09-04
update pmt&digitpmt getter methods and how they are called in outntup…

JamesJieranShen:
…leproc

Fixes issue where default-initialized digitpmt and pmt classes are created in the ntuple files.


JamesJieranShen:
All pmt loops are changed to map-like as well. This was done via an IDE so hopefully everything is caught.

PR #169, opened 2024-09-09
Adding a timestamp variable to the ntuple output. 

MarcFBergevin:
This variable represent the global tracking of time since the start of run in ns. Currently no triggering effect are applied with the exeption that the ULong64_t format forces the variable to be integer.

timestampVariable
triggerTimeVariable

This changes is required for analysis purposes for a few projects.

### DIFF
--- a/macros/validation/electron.mac
+++ b/macros/validation/electron.mac
@@ -13,7 +13,6 @@
 /rat/proc classifychargebalance
 
 /rat/proclast outroot
-/rat/proclast outntuple
 
 /generator/add combo gun:point:poisson
 /generator/vtx/set e- 0.0 0.0 0.0 2.5
@@ -21,4 +20,4 @@
 /generator/rate/set 2.0
 
 
-/run/beamOn 1000
+/run/beamOn 100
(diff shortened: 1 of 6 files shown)


JamesJieranShen:
Hi Marc, thanks for this PR! 
1. Could you fix the formatting issue, and also remove the modifications to the macros?
2. I'm pretty sure the process will segfault when MC is not populated. There does exist a similar entry, `DS::EV::GetUTC`. If this is not set correctly right now, we should do that in `SplitEVDAQ`. I'm not sure if there's a need to store the MC::UTC time (in fact I'm not certain how can this be different than its EV twin). but if this is needed, we can write that value under the MC block in `OutNTuple`.

MarcFBergevin:
Sorry, I thought the message was from Tanner, thanks James! I loaded a clang module that has clang-format so hopefully that works. 

I needed to modify the 3 DAQ that I know people use as the default output for DS::EV:GetUTC did not make much sense.

JamesJieranShen (approved):
Thanks for the changes. These look great!

PR #170, opened 2024-09-09
Extract common waveform analysis utility functions

rpitelka:
Implement an analysis utilities namespace, containing analysis functions from WaveformAnalysis refactored for use in new analysis processors.


JamesJieranShen (changes requested):
Thanks again for your hard work! I added a couple things that we should fix. Of note to other reviewers: this PR have no effect on existing code. Refractoring the waveform analysis class will come as a separate PR.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,185 @@
+#include <RAT/Log.hh>
+#include <RAT/WaveformUtil.hh>
+
+namespace RAT {
+
+namespace WaveformUtil {
+
+std::vector<double> ADCtoVoltage(const std::vector<UShort_t>& adcWaveform, double voltageRes) {
```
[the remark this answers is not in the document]
rpitelka: Sounds good to me, will implement.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,185 @@
...
+  }
+  pedestal /= (pedWindowHigh - pedWindowLow);
+  return pedestal;
+}
+
+std::pair<int, double> FindHighestPeak(const std::vector<double>& voltageWaveform) {
+  /*
+  Calculate the peak (in mV) and the corresponding sample.
+  */
+  double voltagePeak = 999;
```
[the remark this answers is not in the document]
rpitelka: Will do!

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,185 @@
...
+}
+double CalculateTimeCFD(const std::vector<double>& waveform, std::pair<int, double> peak, double constFrac,
+                        int lookBack, double timeStep) {
+  /*
+  Apply constant-fraction discriminator to digitized PMT waveforms.
+  */
+  double voltageThreshold = constFrac * peak.second;
+  int time = GetThresholdCrossing(waveform, peak.first, voltageThreshold, lookBack, timeStep);
+  // Linearly interpolate threshold crossing time
+  double deltav = waveform[time + 1] - waveform[time];
```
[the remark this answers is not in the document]
rpitelka: This is implemented the same as in Interpolate() from WaveformAnalysis, but ok will improve as per suggestion.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  // Make sure not to integrate past the end of the waveform
+  windowEnd = (windowEnd > waveform.size()) ? waveform.size() : windowEnd;
+  // Make sure not to integrate before the waveform starts
+  windowStart = (windowStart < 0) ? 0 : windowStart;
+
+  for (int i = windowStart; i < windowEnd; i++) {
+    double voltage = waveform[i];
+    charge += (-voltage * timeStep) / termOhms;  // in pC
```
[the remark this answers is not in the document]
rpitelka: Will do!

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,185 @@
...
+    // Downward going pulse
+    if (voltage < voltagePeak) {
+      voltagePeak = voltage;
+      samplePeak = i;
+    }
+  }
+  return std::make_pair(samplePeak, voltagePeak);
+}
+
+int GetThresholdCrossing(const std::vector<double>& waveform, int peakSample, double voltageThreshold, int lookBack,
```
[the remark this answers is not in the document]
rpitelka: I like that name change + will improve comments

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,185 @@
...
+    // returned an invalid value
+    if (i == back_window) {
+      thresholdCrossing = 9999;  // Invalid value for bad waveforms
+      break;
+    }
+  }
+  return thresholdCrossing;
+}
+
+std::tuple<int, double, double> GetNCrossings(const std::vector<double>& waveform, double voltageThreshold,
```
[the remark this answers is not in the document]
rpitelka: Sorry should be called like GetCrossingInfo or something like that;  the original function in WaveformAnalysis sets the three values returned by my implmentation, and I don't want to loop over the waveform multiple times to get info we can get in one loop. The implementation for voltageOverThreshold is the same as in the original function.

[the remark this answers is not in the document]
JamesJieranShen: Sounds good. `GetNCrossing(waveform, threshold)->int` sounds like something else that's nice to have by itself as well.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
+#include <RAT/Log.hh>
+#include <RAT/WaveformUtil.hh>
+
+namespace RAT {
+
+namespace WaveformUtil {
+
+const int INVALID = 9999;
```
JamesJieranShen, line 8: can you put this in header? This should be exposed to other namespaces.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+  pedestal /= (pedWindowHigh - pedWindowLow);
+  return pedestal;
+}
+
+std::pair<int, double> FindHighestPeak(const std::vector<double>& voltageWaveform) {
+  /*
+  Calculate the highest peak (in mV) and the corresponding sample.
+  */
+  double voltagePeak = INVALID;
+  int samplePeak = -INVALID;
```
JamesJieranShen, line 77: No... the negative sign here is still a "Magic number". Should this be necessary, you can have SAMPLE_INVALID, ADC_INVALID, and VOLTAGE_INVALID separately.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+                                   int lookBack, double timeStep) {
+  /*
+  Identifies the sample at which threshold crossing occurs before a given peak
+   */
+  int thresholdCrossing = 0;
+  // Make sure we don't scan passed the beginning of the waveform
+  int lb = peakSample - int(lookBack / timeStep);
+  int back_window = (lb > 0) ? lb : 0;
+
+  if (back_window > waveform.size()) {
```
JamesJieranShen, line 99: These checks should be `>=`.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+
+double CalculateTimeCFD(const std::vector<double>& waveform, std::pair<int, double> peak, double constFrac,
+                        int lookBack, double timeStep) {
+  /*
+  Apply constant-fraction discriminator for a given peak
+  */
+  double voltageThreshold = constFrac * peak.second;
+  int time = GetThresholdCrossingBeforePeak(waveform, peak.first, voltageThreshold, lookBack, timeStep);
+  // Linearly interpolate threshold crossing time
+  double deltav = waveform.at(time + 1) - waveform.at(time);
```
JamesJieranShen, line 195: ok -- I think we should just treat this stupid edge case. If `time` is the last element in the vector, don't do the interpolation.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+  }
+  return std::make_tuple(nCrossings, timeOverThreshold, voltageOverThreshold);
+}
+
+double CalculateTimeCFD(const std::vector<double>& waveform, std::pair<int, double> peak, double constFrac,
+                        int lookBack, double timeStep) {
+  /*
+  Apply constant-fraction discriminator for a given peak
+  */
+  double voltageThreshold = constFrac * peak.second;
```
JamesJieranShen, line 192: didn't notice this before, but why is `peak` a pair? This information is already known. let's just do 
`double voltageThreshold = constFrac * waveform.at(peak)`

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+int GetThresholdCrossingBeforePeak(const std::vector<double>& waveform, int peakSample, double voltageThreshold,
+                                   int lookBack, double timeStep) {
+  /*
+  Identifies the sample at which threshold crossing occurs before a given peak
+   */
+  int thresholdCrossing = 0;
+  // Make sure we don't scan passed the beginning of the waveform
+  int lb = peakSample - int(lookBack / timeStep);
+  int back_window = (lb > 0) ? lb : 0;
+
```
JamesJieranShen, line 98: Let's also die if waveform.at(peakSample) is > threshold. I think this will guarantee that the time+1 error won't be emitted in the CFD, so we are at least internally seg-fault-proof.

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+  /*
+  Integrate the digitized waveform over sliding windows to calculate a total charge
+  */
+  double total_charge = 0;
+  int nsliding = int(waveform.size() / slidingWindow);
+
+  for (int i = 0; i < nsliding; i++) {
+    double charge = 0;
+    int sample_start = i * slidingWindow;
+    int sample_end = (i + 1) * slidingWindow;
```
JamesJieranShen, line 237: I think there may be an edge case where sample_end  == waveform.size()?

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+  return pedestal;
+}
+
+double CalculatePedestal(const std::vector<double>& waveform, int pedWindowLow, int pedWindowHigh) {
+  /*
+  Calculate the baseline (in mV) in the window between low - high samples.
+  */
+  double pedestal = 0;
+
+  if (pedWindowLow > waveform.size()) {
```
JamesJieranShen, line 54: should be >=

src/util/src/WaveformUtil.cc
```diff
@@ -0,0 +1,251 @@
...
+  pedWindowHigh = (pedWindowHigh > waveform.size()) ? waveform.size() : pedWindowHigh;
+
+  for (int i = pedWindowLow; i < pedWindowHigh; i++) {
+    pedestal += waveform[i];
+  }
+  pedestal /= (pedWindowHigh - pedWindowLow);
+  return pedestal;
+}
+
+double CalculatePedestal(const std::vector<double>& waveform, int pedWindowLow, int pedWindowHigh) {
```
JamesJieranShen, line 48: This overload is a bit scary. Let's just explicitly make them `CalculatePedestalADC` and `CalculatePedestalmV`.

JamesJieranShen: Oh, and let's use a template variable to combine the logic, instead of copying code.

JamesJieranShen (changes requested):

src/util/src/WaveformUtil.cc
rpitelka: The subsequent loop is `for (int j = sample_start; j < sample_end; j++)`, which is fine if `sample_end == waveform.size()`, no?

JamesJieranShen: Yep, you're right.

JamesJieranShen (approved):
LGTM. @tannerbk I think we are ready to merge.

JamesJieranShen:
> LGTM. @tannerbk I think we are ready to merge.

oh -- and let's squash merge this if that's ok.

PR #174, opened 2024-09-24
Multi-PE waveform analysis chain

JamesJieranShen:
- [x] Add new class: `RAT::DS::WaveformAnalysisResult` and composite it to `RAT::DS::DigitPMT`. This class stores PE information reconstructed from digitized waveforms.
- [x] Write all waveform analysis results to OutNtuple, analogous to MCPEs (skipping the write of figures of merits for now).
- [x] (intermediate) modify current waveform analysis to the new data structures, with fitter name of `Lognormal`
- [x] Split current waveform analysis code into two processors -- `WaveformAnalysisCommon` and `WaveformAnalysisLognormal`. First class should do 0th and 1st order reconstruction. Second class does all lognormal related stuff. 
- [x] Refactor waveform analysis to use `WaveformAnalysisUtil` methods.
- [x] Expose more QOL functions in digitpmt / waveformanalysisresult. For example, time_offset should automatically be applied.


JamesJieranShen:
- [x] Renamed WaveformAnalysisCommon to WaveformPrep, distinguishing it more from other waveform analyzers.
- [] Refactor WaveformAnalysisLognormal into a inherited class -- Only `Configure` and `DoAnalysis` need to be exposed and overwritten by analyzers. Everything else can be nicely contained in the parent class.

JamesJieranShen:
I believe this PR is ready. @rpitelka you might want to take a look at the changes i've made and see if there's objections.

@tannerbk @MarcFBergevin Please review when you have a chance. Thanks!

src/daq/include/RAT/WaveformAnalysisLognormal.hh
```diff
@@ -23,34 +25,30 @@
 #include <RAT/DS/DigitPMT.hh>
 #include <RAT/Digitizer.hh>
 #include <RAT/Processor.hh>
+#include <RAT/WaveformAnalyzerBase.hh>
 #include <vector>
 
+#include "WaveformAnalyzerBase.hh"
```
rpitelka, line 31: Lines 28 and 31 seem redundant.

rpitelka:
Documentation of the changes made to the analysis chain can be found in [these slides](https://nino.lbl.gov/eos/DocDB/cgi/ShowDocument?docid=369).

JamesJieranShen:
Doc updated to EosDocDB 369-v2

src/util/src/WaveformUtil.cc
```diff
@@ -47,13 +47,14 @@ int GetThresholdCrossingBeforePeak(const std::vector<double>& waveform, int peak
   int back_window = (lb > 0) ? lb : 0;
 
   if (back_window >= waveform.size()) {
-    Log::Die("WaveformUtil: Start of lookback window must be before end of waveform.");
+    warn << "WaveformUtil::GetThresholdCrossingBeforePeak: Start of lookback window not before end of waveform."
+         << newline;
   } else if (back_window >= peakSample) {
-    Log::Die("WaveformUtil: Start of lookback window must be before peak.");
+    warn << "WaveformUtil::GetThresholdCrossingBeforePeak: Start of lookback window not before peak." << newline;
```
tannerbk, line 53: This message spams the log because you're running this function for all waveforms, not just for ones with pulses that cross threshold.

src/daq/src/WaveformAnalysisLognormal.cc
```diff
@@ -0,0 +1,147 @@
...
+  ln_fit->SetParameter(2, 0.0);
+  ln_fit->SetParLimits(2, -1.0, 1.0);
+  ln_fit->SetParameter(3, fFitScale);
+  ln_fit->SetParLimits(3, fFitScale - 5.0, fFitScale + 5.0);
+
+  ln_fit->FixParameter(4, fFitShape);
+
+  wfm->Fit("ln_fit", "0QR", "", bf, tf);
+
+  fFittedCharge = ln_fit->GetParameter(0) / fTermOhms;
```
tannerbk, line 123: It would be nice to be able to also return the peak.

src/util/src/WaveformUtil.cc
JamesJieranShen: Resolved in commit [7f2f667](https://github.com/rat-pac/ratpac-two/pull/174/commits/7f2f667a30b94310605ef3354052510a93f572c8).

Despite what the diff here says, this debug msg is actually new (it didn't use to die). So one expects to still see _some_ instances of this message despite the fix, but certainly much less spammy.

PR #176, opened 2024-09-25
Make channel status read from multiple tables

JamesJieranShen:
@llebanowski pinging you here since can't add you as a reviewer;)


llebanowski:
James, thanks for the change and the added functionality.  Looks good.

PR #177, opened 2024-10-01
First implementation of BONSAI for RATPAC2.

MarcFBergevin:
First implementation of BONSAI for RATPAC2. Currently compiles and runs, but random crashes can be observed under certain conditions on my machine. Unsure why private version of software did not have issues. 

Currently does not support processed waveform inputs, only takes QT as inputs.

Missing documentation on how to generate likelihoods.


JamesJieranShen:
Superseded by #251

PR #188, opened 2024-10-11
ChannelStatus: Guarantee a default value

llebanowski:
The previous code does not handle when a run number is specified.

### DIFF
--- a/src/ds/include/RAT/DS/ChannelStatus.hh
+++ b/src/ds/include/RAT/DS/ChannelStatus.hh
@@ -50,17 +50,18 @@ class ChannelStatus : public TObject {
   }
 
   virtual void Load(const PMTInfo* pmtinfo, const std::string index = "") {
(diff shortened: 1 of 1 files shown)


JamesJieranShen:
Hey Logan, thanks for catching this.. I think the culprit of this error is not in the ChannelStatus class. This _should_ work as DBLink::GetI should look through all user planes. I think we should actually fix this here. It appears the conditional logic in `DBLink::Get` is troublesome. I don't think we were supposed to issue a DBNotFoundError here: https://github.com/rat-pac/ratpac-two/blob/5a27cab776903bbbb46c48e85f8b53e28ed73537/src/db/include/RAT/DBLink.hh#L184

JamesJieranShen:
Close this for now -- as discussed on Slack, I will reimplement the table overlap behavior such that we require an experiment to completely overwrite a ratpac2 cable.

PR #201, opened 2024-11-08
Add r14688 timing calibration + arxiv references

JamesJieranShen:
- Add now-published r14688 timing profile measured at Berkeley.
- Add ARXIV reference to the measurement paper.
@hbjamin -- should make sure I'm not crazy.


tannerbk:
Looks good.

PR #202, opened 2024-11-12
Add volume tracking to outntuples

syjeon162:
Add meta branch `volumeCodeMap` and output branch `trackVolume` to `OutNtupleProc`.
`trackVolume` is a 2D array of integers. The volume names corresponding to each integer can be found in `volumeCodeMap`. Volume name is the name of the physical volume as defined in the geometry.
These branches are only created and populated if tracking is on.

The second commit also removes some seemingly unused variables from OutNTupleProc, but if this is poor judgement on my part, please revert it. The first commit is enough for the purpose of this PR.


JamesJieranShen:
Looks good. I think the process tracking works correctly. So i would suggest not to remove that. Can you revert this commit?

JamesJieranShen (approved):
LGTM!

PR #203, opened 2024-11-14
Add Gaussian waveform analysis processor

syjeon162:
Add Gaussian waveform analysis processor, based off of the existing Lognormal waveform analysis processor.

### DIFF
--- a/src/daq/include/RAT/WaveformAnalysisGaussian.hh
+++ b/src/daq/include/RAT/WaveformAnalysisGaussian.hh
@@ -40,8 +40,6 @@ class WaveformAnalysisGaussian : public WaveformAnalyzerBase {
   DBLinkPtr fDigit;
 
   // Analysis constants
-  int fPedWindowLow;
-  int fPedWindowHigh;
   double fFitWindowLow;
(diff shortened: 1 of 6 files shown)


src/daq/include/RAT/WaveformAnalysisGaussian.hh
```diff
@@ -0,0 +1,66 @@
...
+
+  // Fit the digitized waveform using a Gaussian function
+  void FitWaveform(const std::vector<double> &voltWfm);
+
+ protected:
+  // Digitizer settings
+  DBLinkPtr fDigit;
+
+  // Analysis constants
+  int fPedWindowLow;
```
JamesJieranShen, line 43: `fPedWindowxxx` should not be needed since the pedestal calculation should be done in waveformprep. I'm realizing that I left this in the lognormal fit as well. Will fix soon!

syjeon162: this has been fixed!

JamesJieranShen (approved):
LGTM -- difference between lognormal and gaussian is reasonable: 
image
image

PR #204, opened 2024-11-19
Allow user-specified position generator in VertexFile_Gen

mastbaum:
If a position generator is specified in the macro for the `VertexFile_Gen` generator, it will be used for the event vertex position (the location of all primaries). Otherwise, the vertex information from the input file is used.


tannerbk:
LGTM. @MarcFBergevin look OK?

PR #205, opened 2024-11-19
Add RooTracker_Gen to source events from GENIE RooTracker

mastbaum:
This vertex generator works like the `VertexFile_Gen`, but instead of reading RAT ROOT files with `DS::MC` information, it creates MC primaries from the StdHep fields in a GENIE RooTracker file. (This avoids the need to compile RAT and GENIE in the same environment, and to do a RAT-specific conversion step (genie2rat); rather, convert GHEP using GENIE's `gntpc`.)


tannerbk:
I believe an example macro for using this feature was requested.

JamesJieranShen:
From @mastbaum on slack:

>I looked into this, but there are a couple reasons I'd prefer not to:
>1. It needs a GENIE GHEP input ROOT file, so we would need to include one somewhere with the source, and even then the example macro might require some editing to get the path to it correct.
>2. A quick test using the validation geometry crashed, I think for unrelated known reasons (trying to Cherenkov across the world boundary). So we would also need to change that, or add another generic detector geometry just for this example.
>3. The example would really only serve to document the usage of the one rootracker generator line, which is already documented in much more detail in the User Guide.
>So given (3), sorting out (1) and (2) didn't really seem worth it to me.

I personally think this reasoning is fine. Given that this is the only blocker I think we should just go ahead and merge this.

MarcFBergevin (approved):
Based on comments from Conversation, I think we should also merge.

I have compiled the code with no issues.

PR #206, opened 2024-11-21
added charge scale factor in channel status

rl23:


src/core/src/Gsim.cc
```diff
@@ -544,6 +543,7 @@ void Gsim::MakeEvent(const G4Event *g4ev, DS::Root *ds) {
   int numPE = 0;
 
   for (int ipmt = 0; ipmt < hitpmts->GetEntries(); ipmt++) {
+    double chargeScale = 1;
```
JamesJieranShen, line 546: Can we move this down to l. 558? I believe GetChannelStatus should set the default value via the ratdb entry you added.

rl23: done :)

ratdb/CHANNEL_STATUS_DEFAULTS.ratdb
```diff
@@ -12,3 +12,9 @@
   "run_range": [0, 0],
   "default_value": 1,
 }
+{
```
tannerbk, line 15: Can you add a new line here to separate the tables? I think it would make it slightly easier to read.

src/core/src/Gsim.cc
```diff
@@ -555,23 +554,24 @@ void Gsim::MakeEvent(const G4Event *g4ev, DS::Root *ds) {
...
+        AddMCPhoton(rat_mcpmt, a_pmt->GetPhoton(i), NULL, process, chargeScale);
       }
     }
   }
   mc->SetNumPE(numPE);
 }
 
-void Gsim::AddMCPhoton(DS::MCPMT *rat_mcpmt, const GLG4HitPhoton *photon, EventInfo * /*exinfo*/, std::string process) {
+void Gsim::AddMCPhoton(DS::MCPMT *rat_mcpmt, const GLG4HitPhoton *photon, EventInfo * /*exinfo*/, std::string process,
+                       double chargeScale) {
```
tannerbk, line 574: I don't think this should be a new line.

src/core/src/Gsim.cc
```diff
@@ -592,7 +592,7 @@ void Gsim::AddMCPhoton(DS::MCPMT *rat_mcpmt, const GLG4HitPhoton *photon, EventI
   rat_mcphoton->SetCreationTime(photon->GetCreationTime());
 
   rat_mcphoton->SetFrontEndTime(fPMTTime[fPMTInfo->GetModel(rat_mcpmt->GetID())]->PickTime(photon->GetTime()));
-  rat_mcphoton->SetCharge(fPMTCharge[fPMTInfo->GetModel(rat_mcpmt->GetID())]->PickCharge());
+  rat_mcphoton->SetCharge(chargeScale * fPMTCharge[fPMTInfo->GetModel(rat_mcpmt->GetID())]->PickCharge());
```
tannerbk, line 595: Can you add a comment here to clarify what "chargeScale" is doing? For example, "Set the charge for the photoelectron, scaled by an optional calibration parameter 'chargeScale' that is default 1.0".

tannerbk (commented):
Thanks @rl23, I've added some small comments.

llebanowski:
It is probably worth adding this parameter to the meta info in src/io/src/OutNtupleProc.* like the other ChannelStatus parameters.  

It is interesting that the channel charge is implemented in Gsim - is this the only or best place?  

Thanks!

src/core/src/Gsim.cc
rl23: Tried changing back to one line, but we changed back by the formatting check...
`Checking formatting...
./src/core/src/Gsim.cc:573:119: error: code should be clang-formatted [-Wclang-format-violations]
void Gsim::AddMCPhoton(DS::MCPMT *rat_mcpmt, const GLG4HitPhoton *photon, EventInfo * /*exinfo*/, std::string process, double chargeScale) {
`

ratdb/CHANNEL_STATUS_DEFAULTS.ratdb
rl23: done

src/core/src/Gsim.cc
rl23: done

tannerbk:
Looks good to me. @JamesJieranShen?

src/core/include/RAT/Gsim.hh
```diff
@@ -74,15 +74,14 @@ class Gsim : public Producer, G4UserRunAction, G4UserEventAction, G4UserTracking
...
-
   /* Storing optical creation track ID and step */
   void PhotonRecurse(std::vector<int> &PhotonIDs, int trackID, int &parentID, int &firstCreatedID);
   void SetOpticalPhotonIDs(std::string particle_type, int trackID, int parentID);
   std::vector<int> OpticalPhotonIDs;
 
   G4RunManager *theRunManager;
   GLG4DebugMessenger *theDebugMessenger;
-
+  RAT::DS::ChannelStatus ch_status;
```
JamesJieranShen, line 84: This is now unused.

JamesJieranShen (changes requested):

src/core/include/RAT/Gsim.hh
rl23: removed :)

JamesJieranShen (approved):

PR #208, opened 2024-12-02
New waveform analysis class to interpolate waveforms using tapered sinc kernel

YashBezawada:
Added a new WaveformAnalysisSinc class (based on the existing WaveformAnalysis class and the WaveformAnalysisLognormal class) to interpolate waveforms using tapered sinc kernel.


src/daq/src/WaveformAnalysisSinc.cc
```diff
@@ -0,0 +1,110 @@
...
+  for (int j = sample_low; j < sample_high + 1; j++) {
+    fit_wfm.push_back(voltWfm[j]);
+  }
+
+  // Calculating tapered sinc kernel
+  std::vector<double> tsinc_kernel;
+  int N = 8;        // number of interpolated points per data point
+  double T = 30.0;  // tapering constant
+  for (int k = -48; k < 49; k++) {
+    double val = k * 3.1415 / (float)N;
```
JamesJieranShen, line 89: should use c++  style casting (static_cast in this case) instead of c-style casting. ditto for all cases.

src/daq/src/WaveformAnalysisSinc.cc
```diff
@@ -0,0 +1,110 @@
...
+  std::vector<double> fit_wfm;
+  int sample_low = static_cast<int>(floor(bf / fTimeStep));
+  int sample_high = static_cast<int>(floor(tf / fTimeStep));
+  sample_high = std::min(static_cast<int>(voltWfm.size()) - 1, sample_high);
+  for (int j = sample_low; j < sample_high + 1; j++) {
+    fit_wfm.push_back(voltWfm[j]);
+  }
+
+  // Calculating tapered sinc kernel
+  std::vector<double> tsinc_kernel;
```
JamesJieranShen, line 85: I believe this kernel calculation is independent of the input waveform parameters. Meaning that the kernel calculation should be placed in a static function, and should only be computed once, not during every single `DoAnalysis` call. 

Furthermore, currently it seems all kernel parameters are hard-coded. I would suggest one of the following two changes:
1. Expose relevant parameters (for example, ones that adjust the Fourier-Domain bandwidth of the sinc kernel) as a ratdb table entry
2. convert this kernel calculation to a pre-calculated array, placing it in a ratdb entry. 
While option 2 may seem inferior at first glance,  it would actually make this fitter a general "convolution" analyzer. I am not sure how useful this would be since the charge integration is quite simple, but perhaps people would want to run kernels with better time-domain behaviors.

src/daq/src/WaveformAnalysisSinc.cc
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  // Interpolated waveform
+  std::vector<double> interp_wfm = convolve_wfm(fit_wfm, tsinc_kernel);
+  std::pair<int, double> peakSampleVolt = WaveformUtil::FindHighestPeak(interp_wfm);
+
+  fFitPeak = peakSampleVolt.second;
+  fFitTime = bf + peakSampleVolt.first * fTimeStep / (float)N;
+  fFitCharge = 0;
+  for (auto& vlt : interp_wfm) {
```
JamesJieranShen, line 105: do not use `auto`, unless we are looping through maps with pairs being the type.

src/daq/src/WaveformAnalysisSinc.cc
JamesJieranShen: plus, `N` is only used as a float. Perhaps it should just be defined as a float. 

Also, if this kernel is _not_ refactored in a separate scope, please rename the single-letter variables to make them more searchable.

src/daq/src/WaveformAnalysisSinc.cc
```diff
@@ -0,0 +1,110 @@
...
+  // Perform the convolution
+  for (int i = 0; i < n; ++i) {
+    for (int j = 0; j < m; ++j) {
+      result[i + j] += wfm[i] * kernel[j];
+    }
+  }
+  return result;
+}
+
+void WaveformAnalysisSinc::InterpolateWaveform(const std::vector<double>& voltWfm) {
```
JamesJieranShen, line 66: I have a couple issues with the name of this function, as well as accommodating documentations.
1. No "interpolation" is actually performed here. Interpolation would suggest that we are going in between data samples. As far as I can tell, this is not happening. What _is_ happening is a FIR convolution  / filtering. So it should probably named as such.
2. Charge and time calculation is actually done at the end of this method. This is not obvious that this is happening.

src/daq/src/WaveformAnalysisSinc.cc
```diff
@@ -0,0 +1,110 @@
...
+      sinc /= val;
+    tsinc_kernel.push_back(sinc * exp(-pow(k / T, 2)));
+  }
+
+  // Interpolated waveform
+  std::vector<double> interp_wfm = convolve_wfm(fit_wfm, tsinc_kernel);
+  std::pair<int, double> peakSampleVolt = WaveformUtil::FindHighestPeak(interp_wfm);
+
+  fFitPeak = peakSampleVolt.second;
+  fFitTime = bf + peakSampleVolt.first * fTimeStep / (float)N;
```
JamesJieranShen, line 103: My apologies if I'm missing the math here. But this line seems to be indexing an upsampled waveform (therefore dividing sampling period by N) but I don't think any upsampling is done here (the convolution output waveform is N+M-1). 

Again, I could be misunderstanding what you're trying to do. It would be nice if there's a brief explanation of the math being done as a writeup somewhere:)

JamesJieranShen (commented):
Hi Yashwanth, thanks again for the updated PR. I've left some comments on general code restructuring. In general, this looks good! 

I've mentioned in one of the comments, but it would be great if there's a writeup detailing the math being done here for users in the future.

Thanks again!

YashBezawada:
Thanks for the comments, James! I will work on these.

src/daq/src/WaveformAnalysisSinc.cc
YashBezawada: We are interpolating the waveform by convolving the waveform with the tsinc kernel! The convolution introduces 7 new points between two sampled data points. This is done in the `convolve_wfm` function. Sorry for the confusion with the function names. I will update the function names to better reflect what they are doing.

src/daq/src/WaveformAnalysisSinc.cc
YashBezawada: You are correct. I went back and looked at the code. There is an error in my implementation while convolving the kernel and the sampled waveform. I will redo the code and push the changes.

YashBezawada:
Hey @JamesJieranShen, I have updated the code. I am able to run it without any errors. Please let me know if I have to change anything or include any comments, etc.

src/daq/src/WaveformAnalysisSinc.cc
JamesJieranShen: Convolution is now computed correctly. Thanks for the update!

src/daq/include/RAT/WaveformAnalysisSinc.hh
```diff
@@ -0,0 +1,102 @@
...
+  // Digitizer settings
+  DBLinkPtr fDigit;
+
+  // Analysis constants
+  double fFitWindowLow;
+  double fFitWindowHigh;
+  static int fNumInterpPoints;  // number of interpolated points per data point (or the number of points in each lobe of
+                                // the tsinc kernel)
+  static double fTaperingConst;  // tapering constant (determines how fast the kernel decays)
+  static int fNumSincLobes;  // number of sinc lobes to be included in the kernel (determines the length of the kernel)
```
JamesJieranShen, line 73: Why are these values kept static? There's no reason for them to be since they can be individually configured for each instance of the processor.

src/daq/include/RAT/WaveformAnalysisSinc.hh
```diff
@@ -0,0 +1,102 @@
...
+  // Coming from WaveformPrep
+  double fDigitTimeInWindow;
+
+  // Fit variables
+  double fFitTime;
+  double fFitCharge;
+  double fFitPeak;
+
+  static std::vector<double> tsinc_kernel;  // Stores the tsinc kernel
+  static void calculateTSincKernel() {      // Calculates the tsinc kernel
```
JamesJieranShen, line 84: `tsinc_kernel` should not be kept static, same reason as above. If you would like to have the method be kept static, return the computed kernel. 

Also, tsinc_kernel should probably be a `std::array` as they are more performant.

src/daq/src/WaveformAnalysisSinc.cc
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  InterpolateWaveform(voltWfm);
+
+  DS::WaveformAnalysisResult* fit_result = digitpmt->GetOrCreateWaveformAnalysisResult("Sinc");
+  fit_result->AddPE(fFitTime, fFitCharge, {{"peak", fFitPeak}});
+}
+
+std::vector<double> WaveformAnalysisSinc::ConvolveWaveform(const std::vector<double>& wfm) {
+  int wl = wfm.size();  // waveform length
```
JamesJieranShen, line 76: Try pulling the latest version of ratpac2 and see if you get any compile errors -- I think casting `size_t` to `int` will result in a warning since it's unsigned.

JamesJieranShen (approved):
LGTM!

JamesJieranShen:
Thanks @YashBezawada -- we will probably be interested in adding some tests for this fitter. But that could be a separate PR.

src/daq/include/RAT/WaveformAnalysisSinc.hh
YashBezawada: Hey James, thanks for the comments! I have changed the code to not keep the kernel and the other variables as static. 

The size of the kernel is determined by the inputs from the ratdb table (`num_interp_points` and `num_sinc_lobes`). That's why I chose `std::vector` over `std::array`. Is there a better way to do this with `std::array`?

src/daq/src/WaveformAnalysisSinc.cc
YashBezawada: I have changed the code to perform this check differently so that `size_t` is not being casted to an `int`.

src/daq/include/RAT/WaveformAnalysisSinc.hh
JamesJieranShen: I see.. no need to worry about array vs vector here then. My mistake.

PR #214, opened 2024-12-09
rattest QOL update + centroid test

JamesJieranShen:
This PR includes the following QOL updates to rattest:
1. `rattest.py` is now installed to `${CMAKE_INSTALL_PREFIX}/bin`, alongside all other install executables. 
3. tests can now define local ratdb experiments in their own directory. 
4. updated adjacent documentation.
5. test scripts, particularly the root macro, is now included in the clang format checker scope. 
6. fitcentroid has been revived and updated. Previously it was throwing below-cherenkov-threshold electrons in water. I've replaced the detector medium to the default `scintillator` material. Also added a inner detector volume to dramatically speedup simulation time.
7. remove old files that are no longer necessary, per @smnaugle 's suggestion. 
8. Added some basic ANSI coloring to rattest logs. Now the logs show up as below: 
image


.github/workflows/templates/rattests-template.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       run: |
         source /ratpac-setup/env.sh
         source ../ratpac2/ratpac.sh
-        python3 $RATSHARE/python/rattest.py -t test/full/${{ inputs.rattest-name }}
+        python3 rattest -t test/full/${{ inputs.rattest-name }}
```
smnaugle: We could just call rattest directly since the script is properly installed now.

test/full/fitcentroid/fitcentroid.C
```diff
@@ -1,21 +1,62 @@
-void make_plots(TFile *event_file, TTree *T, TFile *out_file)
-{
-  TH1F *hCenterRes = new TH1F("hCenterRes", "20 keVee electrons at the center;Reconstructed X (mm);Events per bin", 40, -800, 800);
-  TProfile *hReconVsRadius = new TProfile("hReconVsRadius", "20 keVee electrons (error bars are RMS);(Radius/437 mm)**3;Reconstructed X - True X (mm)", 5, 0, 1, "s");
-  TProfile *hReconVsEnergy = new TProfile("hReconVsEnergy", "Electrons at center (error bars are RMS);Kinetic energy (keVee);Reconstructed X (mm)", 5, 20, 100,"s");
+// void make_plots(TFile *event_file, TTree *T, TFile *out_file)
```
smnaugle, line 1: Do we want to remove the comments?

test/full/fitcentroid/cylinder/cylinder.geo
```diff
@@ -0,0 +1,58 @@
...
+mother: "", // world volume has no mother
+type: "box",
+size: [20000.0, 20000.0, 20000.0], // mm, half-length
+material: "rock",
+invisible: 1,
+}
+
+{
+name: "GEO",
+index: "water",
```
smnaugle, line 15: Maybe change index to match material? For both inner and outer water geometries.

smnaugle:
My comments are all cosmetic, once @JamesJieranShen pushes the commits that fix the tests (and the tests pass) it looks good to me. I don't know what the removed files under `test` and `python/rattest` were actually for, but they don't seem to be needed so I don't see any harm in getting rid of them. The changes to how ratdb tables can be loaded from within experiment directories is out of my expertise, so someone will have to weigh in on whether or not that's a good change.

JamesJieranShen:
heads up that actually the centroid test might fail -- there's actually a bug right now where the centroid fitter doesn't gracefully handle when total charge of the event is 0. This doesn't lead to a fail every single time, but does with certain seeds.

smnaugle:
> heads up that actually the centroid test might fail -- there's actually a bug right now where the centroid fitter doesn't gracefully handle when total charge of the event is 0. This doesn't lead to a fail every single time, but does with certain seeds.

We should have another PR to fix that bug that we can merge in at the same time as this PR. Are you able to do that @JamesJieranShen ?

tannerbk:
I don't understand the motivation for the events selected to use in the macro. It seems like we're simulating three classes of events, which complicates understanding the output. I'd just simulate central, 5 MeV, isotropic electrons. We could have separate rattests for lower or higher energy events, if we choose.

JamesJieranShen:
> I don't understand the motivation for the generators selected to use in the macro. It seems like we're simulating three classes of events, which complicates understanding the output. I'd just simulate central, 5 MeV, isotropic electrons. We could have separate rattests for lower or higher energy events, if we choose.

are you cool with just keeping the third one:
```
/generator/add combo gun2:point
/generator/vtx/set e- 0 0 0 0 3.5 15.0
/generator/pos/set 0 0 0
```
3.5-15MeV center runs? I'd think residual as a function of energy would be a useful thing.

tannerbk:
Sure, you'll need to update the analysis code accordingly.

JamesJieranShen:
Test updated -- now we simulated 1000 electrons, from 3.5MeV to 15 MeV. We plot the overall distribution of the position distribution, as well as a profile of the residual as a function of energy. The test currently takes quite a while to run (about 15 minutes), due to the increased stats. If this is not necessary I can drop it down to perhaps 100events.

tannerbk:
Looks good to me.

PR #216, opened 2024-12-11
Remove trigger time from the timing

tannerbk:
This removes the addition of the trigger timing in the PMT and digitized PMT timing. There is no reason to add a "trigger time" to the PMT hit-times, particularly since the trigger simulation scheme is currently just a placeholder. Adding the trigger time currently causes nonphysical behavior in the timing (see below), which is resolved by removing it.
image


tannerbk:
This should not be merged as it will cause timing issues for MC events with two triggers. However, I'll leave it open for now as the issue with splitevdaq needs to be addressed.

PR #218, opened 2024-12-13
Fix a bug in waveform util

tannerbk:
This check does not work when `windowStart < 0`. 
` if (windowStart >= waveform.size()) `

### DIFF
--- a/src/util/src/WaveformUtil.cc
+++ b/src/util/src/WaveformUtil.cc
@@ -48,7 +48,7 @@ int GetThresholdCrossingBeforePeak(const std::vector<double>& waveform, int peak
 
   if (back_window >= waveform.size()) {
     debug << "WaveformUtil::GetThresholdCrossingBeforePeak: Start of lookback window not before end of waveform."
-         << newline;
+          << newline;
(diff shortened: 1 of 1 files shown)


src/util/src/WaveformUtil.cc
```diff
@@ -171,16 +171,16 @@ double IntegratePeak(const std::vector<double>& waveform, int peakSample, int in
...
-    charge = INVALID;  // Invalid value for bad waveforms
-    return charge;
-  }
-
   // Make sure not to integrate past the end of the waveform
   windowEnd = (windowEnd > waveform.size()) ? waveform.size() : windowEnd;
   // Make sure not to integrate before the waveform starts
   windowStart = (windowStart < 0) ? 0 : windowStart;
 
+  if (windowStart >= waveform.size()) {
```
llebanowski, line 179: Can have a more useful check now using windowEnd

JamesJieranShen:
@rpitelka take a look?

rpitelka:
Looks good to me!

JamesJieranShen (approved):

PR #220, opened 2025-01-07
Add in quadfitter

smnaugle:
Adds in quadfitter as a rat processor that can be used by analyzers.

TODO: Should add in an experiment agnostic way for users to define what ratdb index to use for quadfitter.

### DIFF
--- a/macros/examples/quadfitter.mac
+++ b/macros/examples/quadfitter.mac
@@ -1,0 +1,26 @@
+# Example macro showing how to use quadfitte
+
+/glg4debug/glg4param omit_muon_processes 1.0
+/glg4debug/glg4param omit_hadronic_processes 1.0
+
+/rat/db/set DETECTOR experiment "Validation"
+/rat/db/set DETECTOR geo_file "Validation/Valid.geo"
+
+/run/initialize
+
+/rat/proc splitevdaq
+/rat/proc count
+/rat/procset update 100
+
+/rat/proc classifychargebalance
+/rat/proc quadfitter
+
+/rat/proclast outntuple
+
+/generator/add combo gun:point:poisson
+/generator/vtx/set e- 0.0 0.0 0.0 2.5
+/generator/pos/set 0.0 0.0 0.0
+/generator/rate/set 2.0
+
+
+/run/beamOn 100


ratdb/FIT_QUAD.ratdb
```diff
@@ -0,0 +1,8 @@
+{
+name: "FIT_QUAD",
```
JamesJieranShen, line 2: This data entry should be placed in FITTER.ratdb, with the index of QUAD, rather than initing its own table.

src/fit/CMakeLists.txt
```diff
@@ -6,7 +6,8 @@ add_library(fit OBJECT
   src/FitPathProc.cc
   src/FitCentroidProc.cc
   src/FitTensorProc.cc
-  src/MiniSim.cc)
+  src/MiniSim.cc
+  src/FitQuadProc.cc)
```
JamesJieranShen, line 10: nitpick: can we switch the order such that the fitters are put together?

src/fit/include/RAT/FitQuadProc.hh
```diff
@@ -0,0 +1,31 @@
...
+
+class FitQuadProc : public Processor {
+ public:
+  FitQuadProc();
+  virtual ~FitQuadProc() {}
+  void BeginOfRun(DS::Run *run);
+  Processor::Result Event(DS::Root *ds, DS::EV *ev);
+
+ private:
+  std::array<uint, 4> ChoosePMTs(uint nhits);
```
JamesJieranShen, line 16: I don't think `uint` is actually a standard c++ type according to the standards, although most compilers to typedef it to unsigned int. Can we just use `unsigned int` instead of `uint` in case we deal with a non-cooperating compiler? Better yet, the contextual type to use here is likely `size_t` since it's indexing into an array.

src/fit/include/RAT/FitQuadProc.hh
```diff
@@ -0,0 +1,31 @@
...
+  std::vector<std::array<uint, 4> > BuildTable(const unsigned int n);
+
+  DS::Run *fRun;
+  DS::PMTInfo *fPMTInfo;
+  uint fNumQuadPoints;
+  uint fMaxQuadPoints;
+  uint fTableCutOff;
+  float fLightSpeed;
+  std::vector<uint> fNumPointsTbl = {0,   0,   0,   0,   1,    5,    15,   35,   70,   126,
```
JamesJieranShen, line 25: should be a const array, not a vector.

src/fit/src/FitQuadProc.cc
```diff
@@ -0,0 +1,221 @@
...
+        double v[3];
+        for (int nt = 0; nt < 3; nt++) v[nt] = g[nt] + h[nt] * tau;
+
+        // Maximum radius (mm) for which we will accept a point.
+        // This is meant to remove wild fits, not directly
+        // restrict them to the physically allowed region.
+        // Indeed, if we cut this off at the PMT sphere surface,
+        // it would introduce a large bias when we took the
+        // median later.  So it is a little larger than that.
+        const double rlimit = 9000;
```
JamesJieranShen, line 191: I think this value needs to be experiment specific?

ratdb/FIT_QUAD.ratdb
```diff
@@ -0,0 +1,8 @@
...
+index: "",
+num_points: 20000,
+max_points: 10000,
+table_cut_off: 20,
+light_speed: 218.794131015,
```
JamesJieranShen, line 7: I _think_ it would make more sense to define this in terms of the "group velocity" refractive index, and divide out the speed of light in vacuum in code.

JamesJieranShen (commented):
Hey Sam, thanks for taking the time to do this! I left a couple comments mostly on code structuring, haven't tested it out yet.

JamesJieranShen:
> TODO: Should add in an experiment agnostic way for users to define what ratdb index to use for quadfitter.

Yes -- I'm working on that;)

smnaugle:
Hi James, I agree with all of your comments aside from the one about where to put the fitter ratdb table. Are we sure that we want to keep the tables for all fitters in the same file? I feel like that may become a bookkeeping nightmare, in SNO+ we allow each fitter to have it's own ratdb file. Each fitter might have a pretty large number of indices depending on the lifetime of the experiment and the complexity of the fitter.

src/fit/src/FitQuadProc.cc
smnaugle: Added a ratdb field for this.

src/fit/include/RAT/FitQuadProc.hh
smnaugle: I used size_t for array sizes and unsigned int for numbers that are supposed to represent `pmt_ids`, that made the most sense to me.

ratdb/FIT_QUAD.ratdb
smnaugle: See my other comment. Happy to debate this.

ratdb/FIT_QUAD.ratdb
smnaugle: Hm, in SNO+ we have used light_speed for a long time in quadfitter and other fitters. I think it ends up being a matter of taste so I'd rather not change it, and this way we don't have to make reference to the speed of light in vacuum.

JamesJieranShen: Sure -- I was thinking we can just use CLHEP's [c_light](https://gitlab.cern.ch/CLHEP/CLHEP/-/blob/develop/Units/Units/PhysicalConstants.h#L58), but sure, keeping it as light_speed is fine with me as well.

src/fit/src/FitQuadProc.cc
JamesJieranShen: Is the ratdb field actually being read into here?

JamesJieranShen:
> Hi @JamesJieranShen, thanks for the comments! I agree with all of them aside from the one about where to put the fitter ratdb table. Are we sure that we want to keep the tables for all fitters in the same file? I feel like that may become a bookkeeping nightmare, in SNO+ we allow each fitter to have it's own ratdb file. Each fitter might have a pretty large number of indices depending on the lifetime of the experiment and the complexity of the fitter.

Hey @smnaugle, I was referring to the json/ratdb internal data-structure. It should be in the same "table" as the other fitter, i.e. have the same "name", but different "index":
```json
{
"name": "Fitter",
"index": "QUAD",
...
}
```
Onto splitting the fitters to different files -- I'm ok with putting the fitter ratdb entry as a separate file for book keeping purposes. However, clearly the current purpose of the FITTER.ratdb file is meant to include all fitter parameters. I would suggest changing the file of that name to FITTENSOR.ratdb if you would like to have a separate FITQUAD.ratdb.

Thanks again!

JamesJieranShen (approved):
LGTM, on the code front. I will leave to @MarcFBergevin and others to see if there's anything else wrong.

smnaugle:
This PR currently crashes when running quadfitter on data. We should hold off on merging until I find the bug.

smnaugle:
OK the previously mentioned crash is now fixed and this PR should be ready to merge in. 

One thing to note though, even when I return `Processor::Result(FAIL);` (line 216 of FitQuadProc.cc), the `validposition_quadfitter` branch in the output ntuples is `true`. Is that the intended behavior? If so, how do I make the fit result invalid? I am currently just returning [0,0,0] as the fit result for failed fits.

src/fit/src/FitQuadProc.cc
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  std::vector<double> pmtx, pmty, pmtz, pmtt;
+  for (int i : ev->GetAllDigitPMTIDs()) {
+    DS::DigitPMT *pmt = ev->GetOrCreateDigitPMT(i);
+    TVector3 pmtpos = fPMTInfo->GetPosition(pmt->GetID());
+    if (pmt->GetDigitizedTime() > 1e6) {
+      continue;
+    }
+    pmtt.push_back(pmt->GetDigitizedTime());
```
tannerbk, line 116: We should have a method for how to select which timing to use in the reconstruction algorithms.

tannerbk:
I think it would make sense to merge #226 first and adapt this PR to use the functionality provided there.

tannerbk:
Also, there should be an example macro updates to include the quad fitter in the example.

smnaugle:
I think this is ready to be merged in again. I should also add a `rattest` too, but I wanted to push these commits ASAP so you can reprocess data with it Tanner. I can submit another PR with the `rattests` soon.

PR #222, opened 2025-01-08
Specify experiment binaries with rattests

smnaugle:
Add in command line argument to specify which experiment binary should be used with the rattests. This is useful because it allows for downstream experiments to use the rattests framework while running their tests using their own experiment binary.

### DIFF
--- a/python/rattest.py
+++ b/python/rattest.py
@@ -36,7 +36,7 @@ parser.add_argument('-t', '--text-only',
 
 parser.add_argument('-e', '--experiment',
                     type=str, dest='experiment_bin', default=None,
-                    help='Absolute path to experiment binary to run test with.')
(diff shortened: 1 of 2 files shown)


python/rattest.py
```diff
@@ -34,6 +34,10 @@
                     action='store_false', dest='web', default=True,
                     help='Do not open web pages with plots.')
 
+parser.add_argument('-e', '--experiment',
+                    type=str, dest='experiment_bin', default=None,
+                    help='Absolute path to experiment binary to run test with.')
```
JamesJieranShen, line 39: can we specify here that default is the `rat` binary? Thanks!

JamesJieranShen (approved):
LGTM, other than the small change here. Thanks Sam!

python/rattest.py
smnaugle: good idea, just changed that.

PR #224, opened 2025-01-15
Add a forced trigger option that ignores the light in the detector

tannerbk:
A simple trigger that just forces an event to trigger and sets the trigger time to zero.

### DIFF
--- a/src/daq/src/ForcedTriggerProc.cc
+++ b/src/daq/src/ForcedTriggerProc.cc
@@ -1,4 +1,4 @@
-/* 
+/*
 Forced trigger causes a trigger regardless of the total number of PMTs hit
  */
 #include <G4ThreeVector.hh>
@@ -12,7 +12,6 @@ namespace RAT {
 
 ForcedTriggerProc::ForcedTriggerProc() : Processor("forcedtrigger") {
   // Trigger Specifications
-
   ldaq = DB::Get()->GetLink("DAQ", "ForcedTrigger");
   fEventCounter = 0;
   fDigitizerType = ldaq->GetS("digitizer_name");
@@ -33,7 +32,6 @@ void ForcedTriggerProc::BeginOfRun(DS::Run *run) {
 }
 
 Processor::Result ForcedTriggerProc::DSEvent(DS::Root *ds) {
-
   DS::MC *mc = ds->GetMC();
   DS::Run *run = DS::RunStore::Get()->GetRun(ds);
   DS::PMTInfo *pmtinfo = run->GetPMTInfo();
--- a/src/daq/src/ForcedTriggerProc.cc
+++ b/src/daq/src/ForcedTriggerProc.cc
@@ -10,7 +10,9 @@ Forced trigger causes a trigger regardless of the total number of PMTs hit
 
 namespace RAT {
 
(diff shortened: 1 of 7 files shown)


tannerbk:
@JamesJieranShen maybe you can have a quick look at well since this was your suggestion.

src/daq/src/ForcedTriggerProc.cc
```diff
@@ -0,0 +1,76 @@
...
+#include <RAT/DS/MCPMT.hh>
+#include <RAT/DS/PMT.hh>
+#include <RAT/DS/RunStore.hh>
+#include <RAT/ForcedTriggerProc.hh>
+
+namespace RAT {
+
+ForcedTriggerProc::ForcedTriggerProc() : Processor("forcedtrigger") {
+  // Trigger Specifications
+  ldaq = DB::Get()->GetLink("DAQ", "ForcedTrigger");
```
JamesJieranShen, line 15: These db entries technically can be run-depednent, so I would suggest moving them to `BeginOfRun`

src/daq/src/ForcedTriggerProc.cc
```diff
@@ -0,0 +1,76 @@
...
+
+  fDigitizer = new Digitizer(fDigitizerType);
+}
+
+void ForcedTriggerProc::BeginOfRun(DS::Run *run) {
+  if (fDigitize) {
+    DS::PMTInfo *pmtinfo = run->GetPMTInfo();
+    const size_t numModels = pmtinfo->GetModelCount();
+    for (size_t i = 0; i < numModels; i++) {
+      const std::string modelName = pmtinfo->GetModelName(i);
```
JamesJieranShen, line 28: nitpick: this should be a `const std::string &` to avoid additional copying.

src/daq/src/ForcedTriggerProc.cc
```diff
@@ -0,0 +1,76 @@
...
+  if (ds->ExistEV()) ds->PruneEV();
+
+  DS::EV *ev = ds->AddNewEV();
+  ev->SetID(fEventCounter++);
+  ev->SetCalibratedTriggerTime(0.0);
+  ev->SetUTC(mc->GetUTC());
+  double totalEVCharge = 0;
+  // Loop over the mcpmts and fill the pmt branch assuming we've triggered
+  for (int imcpmt = 0; imcpmt < mc->GetMCPMTCount(); imcpmt++) {
+    DS::MCPMT *mcpmt = mc->GetMCPMT(imcpmt);
```
JamesJieranShen, line 48: super nitpick: everything in MC really should be const..

JamesJieranShen (approved):
LGTM in general -- i think all issues i've pointed out are also present in splitevdaq, so no major issues since we've been running that for a while without major issues:)

src/daq/src/ForcedTriggerProc.cc
tannerbk: Moved for both this processor and splitevdaq

azummo:
I simulated a couple of electrons with splitevdaq and forcedtrigger and it seems to behave as expected (every event triggers with triggertime=0 for forcedtrigger). The code looks good to me as well.

PR #226, opened 2025-01-16
Fitter interface

JamesJieranShen:
Add a new class -- `FitterInputHandler`, which returns charge and timing information of the event to fitters. This removes the need for fitters to deal with multiple types of PMT information, from DigitPMT/PMT/Waveformanalyzers. 

Example of how to use this is provided for the centroid fitter. Running the processor as-is uses DS::PMT information, as this is the default for most standard RAT experiments. To switch to Lognormal waveform analysis, use the following example snippet:

```
/rat/db/set FIT_COMMON mode 2 # 0 is DS::PMT, 1 is DigitPMT, 2 is waveformAna
/rat/db/set FIT_COMMON waveform_analyzer "Lognormal" # analyzer name
```

Should solve some structural issues with #220 #177

### DIFF
--- a/src/fit/include/RAT/FitterInputHandler.hh
+++ b/src/fit/include/RAT/FitterInputHandler.hh
@@ -1,0 +1,214 @@
+/** @class FitterInputHandler
+ * Interface for handling the input stream of a standard fitter.
+ *
+ * @author James Shen <jierans@sas.upenn.edu>
+ *
+ * By default, the handler gets configured to one stream of input. This "steram"
+ * can either be the DS::PMTs, DS::DigitPMTs, waveform analysis results, etc.
+ *
+ * */
+
+#ifndef __RAT_FitterInputHandler__
+#define __RAT_FitterInputHandler__
+
+#include <RAT/DB.hh>
+#include <RAT/DS/EV.hh>
+#include <RAT/Log.hh>
+#include <algorithm>
+#include <string>
+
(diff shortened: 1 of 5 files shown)


tannerbk:
Looks like a nice idea, I will review soon.

src/fit/include/RAT/FitterInputHandler.hh
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * @brief Return the (approximate) number of hits registered on a PMT.
+   * Behavior is different depending on the mode.
+   *  If mode is set to kPMT, always return 1 since no information about nhit is given.
+   *  If mode is set to kDigitPMT, return the number of times that the waveform crosses threshold.
+   *  If mode is set to kWaveformAnalysis, return the number of hits created by the analyzer.
+   *
+   * @param id PMT ID.
+   */
+  unsigned int GetNhits(Int_t id) {
```
tannerbk, line 187: This is confusingly names. "Nhits" refers to the number of PMT hits in an event, not the number of "hits" on a PMT. This could be more aptly named "GetNPEs".

Additionally, you could add a method for nhits that returns the length of the appropriate hit PMT list.

JamesJieranShen: Both implemented -- the old GetNHits is now called GetNPEs. There's now a new GetNHits method.

tannerbk:
LGTM

PR #231, opened 2025-01-29
Add matched filter waveform analysis processors.

rpitelka:
Add single PE matched filter processor and l1 deconvolution (sparse regression) processors. 

TO-DO
- generalize template
   - currently template values can be stored in DIGITIZER_ANALYSIS ratdb table
   - need to allow for per PMT model type templates
   - templates could go in PMTPULSES data_driven tables, but analysis processor framework doesn't have access (I think)
- fix l1 deconvolution charge normalization
- make sure adding Eigen dependency is ok
- write rattest for processors
- add iterative matched filter processor


rpitelka:
Added Eigen headers directly to rat instead of incorporating into ratpac-setup, and also renamed src/external to src/contrib, both as per James's suggestions.

PR #234, opened 2025-02-10
Update DAQ processor documentation

WilfS:
There was next to no information on the DAQ processors included in ratpac. These are very useful for liquido so I've gone through the source code to get an idea of what each one does. So no one else has to do this I've added some information in the documentation on each processor. 

What I've written would definitely benefit from a look over from whoever wrote the processors.

### DIFF
--- a/doc/users_guide/processors.rst
+++ b/doc/users_guide/processors.rst
@@ -24,14 +24,21 @@ None.
 
 lesssimpledaq
 `````````````
-Here, the timing and charge information of each PMT is estimated the same way as
-simpledaq, but a minimal trigger is also simulated based on a trigger threshold
-and a trigger time window. The trigger threshold is the minimum number of PMTs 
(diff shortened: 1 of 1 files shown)


doc/users_guide/processors.rst
```diff
@@ -6,9 +6,11 @@ simpledaq
 `````````
 The SimpleDAQ processor simulates a minimal data acquisition system.  The time
 of each PMT hit is the time of the first photon hit plus the timing
-distribution of the appropriate PMT, and the charge collected at each PMT is
-just the sum of all charge deposited at the anode, regardless of time.  All PMT
-hits are packed into a single event.
+distribution of the appropriate PMT (i.e. the "frontEndTime" of the first 
+photon), and the charge collected at each PMT isjust the sum of all charge
```
tannerbk, line 10: small typo "isjust"

tannerbk (commented):
Looks good, thank you!

PR #235, opened 2025-02-13
Move pre-begin configuration to a dedicated Configrue() method for better downstream extension

JamesJieranShen:
Some structural changes to the core Rat executable. Previously pre-event-loop configuration are scattered between the ctor and in the Begin() method. This is all moved to a dedicated Configure() method, called by the Begin() method prior to the main event loop. This allows downstream experiment to override this method in a more flexible way than before. Concretely, this solves the issue that downstream processors do not have access to the ratdb tables.

Note that this change is effectively trivial -- `tools/rat.cpp` calls the constructor and Begin() method back-to-back. The only complexity comes from how the rat class is extended in downstream experiments like EosSimulations.

### DIFF
--- a/src/ratbase/include/RAT/Rat.hh
+++ b/src/ratbase/include/RAT/Rat.hh
@@ -8,6 +8,7 @@
 #include <RAT/ProducerBlock.hh>
 #include <RAT/RatMessenger.hh>
 #include <set>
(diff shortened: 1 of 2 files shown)


tannerbk:
LGTM.

PR #236, opened 2025-02-13
Add stopped pion neutrino source flux as option for generators to sample

hbjamin:
Some things to consider before this is merged 

- In `stpi.ratdb` the `fNuType` and `fNuFlavor` are the same, because there is no processes to distinguish between like `hep` and `pp` in `solar.ratdb`.  This means the macro is currently run with `/generator/vtx/set [xdir] [ydir] [zdir] STPI:nue:nue` instead of `/generator/vtx/set [xdir] [ydir] [zdir] SOLAR:hep:nue`. 

- Two of the generators `src/gen/src/ESgen.cc` and `src/gen/src/CCgen.cc` explicitly handle the decision of sampling the spectra from either `solar.ratdb` or `ibd.ratdb`, so the option to use `stpi.ratdb` has also been added.


JamesJieranShen:
Hi @hbjamin , great work! Couple of comments:
1. If there's no specific purpose in keeping code block commented-out, can you please remove it?
2. I don't think having both nutype and nuflavor encode the flavor is a good idea -- surely this would cause confusion down the line. May I suggest setting up scaffolding to have the nutype encode timing information? For now, perhaps nutype can be set to `full` to denote time-integrated distribution. Down the line, we may have `SNS_1us`, `SNS_2us`, etc. 
3. Can we document where the `flux` is currently coming from? For example, we should say `flux assumed to be X m away from SNS target`. We should probably also mention in doc somewhere that oscillation is not considered.

JamesJieranShen:
I chatted with Ben about the comments -- for now the ES/CC generator is rather flawed in flavor support. What should really happen is that each flux should have different flavor components, and one can choose in macro which flux/type and flavor to support. This should happen later down the line though and not within this PR. I do think we should assess the other comments before this gets merged in.

hbjamin:
I agree with @JamesJieranShen. Ideally the `fNuFlavor` and `fNuType` are not the same. But currently each table only has one flux per index and the neutrino flavor is declared outside the table, so I used the table index to label the different fluxes by flavor. 

1. Fixed

2/3. Ideally the names of the tables would be `stpi`, the indices could be different distances from the source `10m`, `20m`, `40m`, or different timing cuts on the beam pulse, so all neutrinos that arrive within the first `2us`, `5us`, or the full `10us`. Then can choose between `spec_flux_nue`, `spec_flux_numu`, `spec_flux_numubar` in `ESgen.cc` and `CCgen.cc` using the nu flavor specified. 

I can go ahead and start making these changes, but should first decide if the indices should be distances from the source or timing cuts.

JamesJieranShen:
After some offline discussion, this PR cannot currently be merged for the following couple of reasons:
1. We should systematically allow all flavors for the ES generator. A lot of assumption is being made in the ES generator at the moment regarding what flavor each spectrum takes. This is not scalable at all in the context of an open-source rat-pac.
2. The CCGenerator is seemingly only for Lithium 7? If so, we should certainly rename that generator. A lot needs to be checked on what is going on with this generator.

I think the flux can be merged in the current stage. I will do so and continue the work on the generator in another PR.

JamesJieranShen:
After some offline discussion, there are still some issues remaining with the current implementation of this PR.
1. The usage of ES/CC generators are not generator. The current code makes a lot of assumptions on what flavor each flux should be. This is certainly not scalable in the context of RAT-PAC. 
2. The CC Generator appears to be only for one element -- Li7. it seems there's a bunch of similar generators for other elements. We should probably combine them into some generalized CC generator.

However, the currently implementation _works_ despite all the changes needed in the generator. I propose this PR to be merged as-is, and that we continue the generator work down the line.

PR #239, opened 2025-03-05
Changing quadfit result for failed fits.

smnaugle:
The quadfit processor will now return fits with [x,y,z,t] = -9999 rather than 0 when a fit fails.

### DIFF
--- a/src/fit/src/FitQuadProc.cc
+++ b/src/fit/src/FitQuadProc.cc
@@ -118,8 +118,8 @@ Processor::Result FitQuadProc::Event(DS::Root *ds, DS::EV *ev) {
   DS::FitResult *fit = new DS::FitResult("quadfitter");
   fit->SetValidEnergy(false);
   fit->SetValidDirection(false);
-  fit->SetPosition(TVector3(0, 0, 0));
-  fit->SetTime(0);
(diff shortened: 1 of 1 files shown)


tannerbk:
LGTM, but is there some global fit invalid value or anything like that? Its probably better not to hard-code this into all of the fitters.

JamesJieranShen:
@smnaugle should we set validPosition to False?

JamesJieranShen (approved):
LGTM!

smnaugle:
> LGTM, but is there some global fit invalid value or anything like that? Its probably better not to hard-code this into all of the fitters.

I couldn't find anything after searching briefly. Should we add one in a separate PR, maybe in `Processor.hh`?

JamesJieranShen:
> > LGTM, but is there some global fit invalid value or anything like that? Its probably better not to hard-code this into all of the fitters.
> 
> I couldn't find anything after searching briefly. Should we add one in a separate PR, maybe in `Processor.hh`?

this is in the works -- we should have a RAT::DS::INVALID=-9999 somewhere in the code base, and do a grep for all the -9999s and replace that. There's a bunch scattered in the waveform analysis world as well. I haven't gotten the chance to do that;)

PR #240, opened 2025-03-05
Pulls laserball spectrum, measured on 01/27/25

srig19:
VertexGen_PhotonBomb.hh has two new private variables, `fDist` and `fWavelength`.
VertexGen_PhotonBomb.cc sets `fDist` to the rat parameter `include_dist` in `LBSPECTRA.ratdb` (see other PR) and samples from the distribution if `fDist` is true. The sampling is contained within a new function, `pickWavelength`.

### DIFF
--- a/ratdb/OPTICS.ratdb
+++ b/ratdb/OPTICS.ratdb
@@ -393,6 +393,48 @@ PROPERTY_LIST: ["RINDEX", "ABSLENGTH", ]
 
 {
 name: "OPTICS",
+index: "borosilicate_glass",
+valid_begin : [0, 0],
+valid_end : [0, 0],
+surface: 1,
+finish: "ground",
+model: "glisur",
+type: "dielectric_dielectric",
+polish: 0.9,
+RINDEX_option: "wavelength",
+// Refractive index of Schott BK7 from https://refractiveindex.info/?shelf=popular_glass&book=BK7&page=SCHOTT
+// There is no data below 300 nm, so the Sellmeier formulas extrapolated below using an exponential fit.
+RINDEX_value1: [60.0, 100.0, 200.0, 300.0, 400.0, 500.0, 600.0, 700.0, 800.0, ],
+RINDEX_value2: [1.707, 1.662, 1.589, 1.551, 1.531, 1.521, 1.516, 1.513, 1.512, ],
+ABSLENGTH_option: "wavelength",
+ABSLENGTH_value1: [60.0, 200.0, 300.0, 330.0, 500.0, 600.0, 770.0, 800.0, ],
+ABSLENGTH_value2: [0.1e-3, 0.1e-3, 0.1e-3, 1.0e3, 2.0e3, 1.0e3, 0.5e3, 0.1e-3, ],
+PROPERTY_LIST: ["RINDEX", "ABSLENGTH", ]
+}
+
+{
+name: "OPTICS",
+index: "hamamatsu_borosilicate_glass",
+valid_begin : [0, 0],
+valid_end : [0, 0],
+surface: 1,
+finish: "ground",
+model: "glisur",
+type: "dielectric_dielectric",
+polish: 0.9,
+// Refractive index of Schott BK7 from https://refractiveindex.info/?shelf=popular_glass&book=BK7&page=SCHOTT
+// There is no data below 300 nm, so the Sellmeier formulas extrapolated below using an exponential fit.
+RINDEX_option: "wavelength",
+RINDEX_value1: [60.0, 100.0, 200.0, 300.0, 400.0, 500.0, 600.0, 700.0, 800.0, ],
+RINDEX_value2: [1.707, 1.662, 1.589, 1.551, 1.531, 1.521, 1.516, 1.513, 1.512, ],
+ABSLENGTH_option: "wavelength",
+ABSLENGTH_value1: [60.0, 200.0, 300.0, 330.0, 500.0, 600.0, 770.0, 800.0, ],
+ABSLENGTH_value2: [0.1e-3, 0.1e-3, 0.1e-3, 1.0e3, 2.0e3, 1.0e3, 0.5e3, 0.1e-3, ],
+PROPERTY_LIST: ["RINDEX", "ABSLENGTH", ]
+}
+
+{
+name: "OPTICS",
 index: "mirror",
 valid_begin : [0, 0],
 valid_end : [0, 0],


src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -15,7 +15,7 @@
 namespace RAT {
 VertexGen_PhotonBomb::VertexGen_PhotonBomb(const char *arg_dbname) : GLG4VertexGen(arg_dbname) {
   fOpticalPhoton = G4ParticleTable::GetParticleTable()->FindParticle("opticalphoton");
-  SetState("1 400");  // one photon per event, 400 nm
+  SetState("1 401");  // one photon per event, 400 nm, don't use distribution
```
tannerbk, line 18: Is there a reason you changed the default value here?

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -25,18 +25,57 @@ VertexGen_PhotonBomb::VertexGen_PhotonBomb(const char *arg_dbname) : GLG4VertexG
 VertexGen_PhotonBomb::~VertexGen_PhotonBomb() { delete fRndmEnergy; }
 
 void VertexGen_PhotonBomb::GeneratePrimaryVertex(G4Event *event, G4ThreeVector &dx, G4double dt) {
-  for (int i = 0; i < fNumPhotons; i++) {
     // Pick direction isotropically
-    G4ThreeVector mom;
-    double theta = acos(2.0 * G4UniformRand() - 1.0);
-    double phi = 2.0 * G4UniformRand() * CLHEP::pi;
+  G4ThreeVector mom;
+  double theta = acos(2.0 * G4UniformRand() - 1.0);
```
tannerbk, line 30: Shouldn't theta and phi remain in the loop over fNumPhotons?

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -25,18 +25,57 @@ VertexGen_PhotonBomb::VertexGen_PhotonBomb(const char *arg_dbname) : GLG4VertexG
...
+  double wavelength;
+  double evWavelength;
+  double evEnergy; // Energy pulled from the spectrum for the event. Ensures all photons within the same event have the same energy
+
+  // Use fixed energy unless spectrum was provided, fDist refers to spectrum sampled on 01/27/25
+  double energy;
+
+  if(fDist){
+    try {
+      //info << std::to_string((int)fWavelength);
```
tannerbk, line 41: You can remove all of these commented out info and debug statements.

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -25,18 +25,57 @@ VertexGen_PhotonBomb::VertexGen_PhotonBomb(const char *arg_dbname) : GLG4VertexG
...
+      DBLinkPtr spectradb = DB::Get()->GetLink("LBSPECTRA", std::to_string((int)fWavelength));
+      double wl_average = spectradb->GetD("avg_wl");
+      debug << "Successfully pulled the average wavelength - " << std::to_string(wl_average) << newline;
```
tannerbk, line 44: I would remove this, this will spam the log when using debug, since it prints for every single photon.

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -25,18 +25,57 @@ VertexGen_PhotonBomb::VertexGen_PhotonBomb(const char *arg_dbname) : GLG4VertexG
...
+  double evWavelength;
+  double evEnergy; // Energy pulled from the spectrum for the event. Ensures all photons within the same event have the same energy
+
+  // Use fixed energy unless spectrum was provided, fDist refers to spectrum sampled on 01/27/25
+  double energy;
+
+  if(fDist){
+    try {
+      //info << std::to_string((int)fWavelength);
+      DBLinkPtr spectradb = DB::Get()->GetLink("LBSPECTRA", std::to_string((int)fWavelength));
```
tannerbk, line 42: You should add an example spectrum to ratdb. Also, I would name this something more generic, like "WAVELENGTH_SPECTRUM".

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -25,18 +25,57 @@ VertexGen_PhotonBomb::VertexGen_PhotonBomb(const char *arg_dbname) : GLG4VertexG
...
+      double wl_average = spectradb->GetD("avg_wl");
+      debug << "Successfully pulled the average wavelength - " << std::to_string(wl_average) << newline;
+      evWavelength = wl_average;
+      try{
+        std::vector<double> wls = spectradb->GetDArray("dist_wl");
+        std::vector<double> wl_probs = spectradb->GetDArray("dist_wl_intensity");
+        debug << "Found the distribution for " << std::to_string((int)fWavelength) <<" and pulling values from here" << newline;
+        if (wls.size() != wl_probs.size()) Log::Die("PDFPMTCharge: charge and probability arrays of different length");
+        evWavelength = pickWavelength(wls, wl_probs, wl_average);
```
tannerbk, line 51: Don't you want to pick the wavelength in the loop over fNumPhotons?

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -51,6 +90,7 @@ void VertexGen_PhotonBomb::GeneratePrimaryVertex(G4Event *event, G4ThreeVector &
     G4ThreeVector e2 = mom.unit().cross(e1);
     G4ThreeVector pol = e1 * cos(phi) + e2 * sin(phi);
     photon->SetPolarization(pol.x(), pol.y(), pol.z());
+    debug << "Polarization successfully set for photon " << std::to_string(i) << newline;
```
tannerbk, line 93: I would remove this debug statement.

src/gen/src/VertexGen_PhotonBomb.cc
srig19: It was relevant to an alternate method that I forgot to revert. I'll change it back to the default.

src/gen/src/VertexGen_PhotonBomb.cc
srig19: Wouldn't we want all the photons for each events to be of the same wavelength? Since they originate from the same laser pulse.

src/gen/src/VertexGen_PhotonBomb.cc
srig19: Made one, going to include it in the next update

src/gen/src/VertexGen_PhotonBomb.cc
srig19: I think it would still be good to have the debug statement so it's easy to spot this error in the log file. I'll set it up in the `SetState` function instead so the debug is only called twice (once for default, once for user values) rather than for each event.

src/gen/src/VertexGen_PhotonBomb.cc
tannerbk: I think it makes more sense for every photon to draw from the distribution, rather than every event. This will work more generally for any situation you would want to use the generator, and even in the specific use-case for Eos, I think a single laser pulse probably consists of photons with a wavelength distribution rather than a single wavelength.

PR #241, opened 2025-03-06
Updated a check to prevent sinc processor from crashing

YashBezawada:
The `fDigitTimeInWindow` variable is returning `-nan` value. Updated a check to prevent the sinc processor from crashing because of this. Incorporated the suggestions from James.


JamesJieranShen (approved):
LGTM. Thanks!

PR #244, opened 2025-03-17
Add optical properties for borosilicate glasses

llebanowski:
Two borosilicate glass materials were defined with no associated optical properties.  The optical definitions are copied from "glass" except for the refractive index, which is taken to be that of Schott BK7 from https://refractiveindex.info/?shelf=popular_glass&book=BK7&page=SCHOTT

The Sellmeier formula represents the data very well, except there is no data below 300 nm.  Therefore the index is extrapolated below using an exponential fit:
image


tannerbk:
LGTM.

Issue #243, opened 2025-03-14
Data driven PMT pulses not functional


WilfS:
I noticed that, when using data driven pmt pulses with the waveformPrep processor, the output waveforms were always flat. I think the issue is on [L52](https://github.com/rat-pac/ratpac-two/blob/56c153cd921279554196a013bc86ac6a842855f3/src/daq/src/PMTPulse.cc#L52) of PMTPulse.cc, where the extracted pulse height isn't assigned to val, so the PMT voltage is always zero.

Should be a quick fix. I'll submit a pull request with the fix, but opening an issue for now in case I've misinterpreted something.

PR #246, opened 2025-03-17
Fix bug affecting data driven pmt pulses

WilfS:
The voltage readout from data driven pmt pulses was not being assigned to the pmtpulse output correctly. This should fix it.

Addresses issue https://github.com/rat-pac/ratpac-two/issues/243


tannerbk:
LGTM.

Issue #245, opened 2025-03-17
Interpolation of waveform threshold crossing can return infinity


WilfS:
In WaveformUtils, linear interpolation is used to get a precise estimation of the crossing time for waveforms over the voltage threshold. If the voltage difference between the two samples crossing the threshold is zero, the linear interpolation returns an infinity.

I've seen this behaviour using data driven pulse shapes in our liquido simulations. My solution to fix this has been to set the time addition from the linear interpolation to zero if the voltage difference is zero. I'll submit a PR with my fix.

I'm not exactly sure why it is possible for the voltage difference at the threshold crossing can be zero, considering the voltage has to change in order to cross the threshold in the first place.

PR #247, opened 2025-03-17
Fix bug in threshold crossing interpolation

WilfS:
The linear interpolation used to get precise threshold crossings of waveforms could return an infinite time if the voltage difference is zero. I've implemented a fix by setting the interpolation to zero if the voltage difference is zero.

Addresses issue https://github.com/rat-pac/ratpac-two/issues/245

### DIFF
--- a/src/util/src/WaveformUtil.cc
+++ b/src/util/src/WaveformUtil.cc
@@ -157,7 +157,11 @@ double CalculateTimeCFD(const std::vector<double>& waveform, int peakSample, int
   double dt = 0;
   if (time + 1 < static_cast<int>(waveform.size())) {
     double deltav = waveform.at(time + 1) - waveform.at(time);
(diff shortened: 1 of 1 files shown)


JamesJieranShen (approved):
LGTM -- thanks for the fix!

JamesJieranShen:
Closes #245

tannerbk:
Hmm, does dt=0 make sense for the interpolation here, or should some midpoint between the two samples be selected?

JamesJieranShen:
> Hmm, does dt=0 make sense for the interpolation here, or should some midpoint between the two samples be selected?

Presumably this condition only occurs when the threshold equals to the sample on the left -- so threshold crossing happened exactly at the earlier sample. Otherwise there would have been an interpolation call between the left sample and the one before it.

PR #249, opened 2025-03-27
Change all PMT glass types

llebanowski:
This PR swaps PMT glass type definitions for all PMTs, following the update in [PR 244](https://github.com/rat-pac/ratpac-two/pull/244).  A before and after comparison will be provided later.

One PMT uses a quartz window, and so the optical properties of quartz are also updated here.  A visualization of the refractive index: 
image


tannerbk:
Can you confirm a negligible impact on nhit and timing?

llebanowski:
Comparison of Before and After changing validPMT glass_material from “glass” to “borosilicate_glass” (changes refractive index).
10,000 events in Valid.geo using macro validation/electron.mac

No discernible impact observed:

image
image
image

PR #250, opened 2025-03-27
Laserball spectra

srig19:


src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -25,18 +25,42 @@ VertexGen_PhotonBomb::VertexGen_PhotonBomb(const char *arg_dbname) : GLG4VertexG
...
-    double energy;
-    if (fRndmEnergy)
+    theta = acos(2.0 * G4UniformRand() - 1.0);
+    phi = 2.0 * G4UniformRand() * CLHEP::pi;
+    if (fRndmEnergy) {
       energy = fMinEnergy + (fMaxEnergy - fMinEnergy) * fRndmEnergy->shoot();
-    else
+    } else if (fDist) {
+      try {
+        DBLinkPtr spectradb = DB::Get()->GetLink("WLSPECTRA", std::to_string((int)fWavelength));
```
JamesJieranShen, line 44: I'm a bit confused on how one is supposed to use the spectrum here. if the name `WLSPECTRA` is hard coded into the script, how would one invoke `LBSPECTRA`?

JamesJieranShen (changes requested):
Hi Yash,

Thanks again for putting in this patch. It would be really nice to have this. I have a couple issues with the current implementation.
1. As far as I can tell, the choice of whether one is choosing to use a spectrum or just simulating monoenergetic photons is only controlled  by a RADB entry. I don't think this is transparent enough, since a user can be running a macro that does something completely unexpected if they don't closely examine the content of their RATDB directory.
2. While I understand where you are coming from, I strongly disagree with the approach you are taking by casting a user input wavelength from int to a string and use it as an index. This can be extremely dangerous, considering JSON (RATDB) is an untyped language and a lot of things can go wrong with the conversion. I also think that this is, again, too much "magic" -- we should make the users clearly specify the spectrum they would like to use, should they choose to use it. 
3. There's actually a potential issue with the way you are specifying a distribution here. As @gmwendel and many others have pointed out previously, since the spectrum is a partial distribution, it is ambiguious if the bins should be considered to be evenly spaced in wavelength or energy. For other similar things (particularly in the scintillation model), we actually specify an `option` (either `wavelength` or `dy_dwavelength`) to account for this. I think we should use a similar approach here. You can refer to the implementation for the scintillation model to see what we should do.


That being said, I propose the following big-picture changes:
- remove the `param` entry to ratdb. 
- For every spectrum, the table name should be `LBSPECTRA` and the index should explicitly be a string, e.g. `400nm`. 
- The photonbomb interface in a macro should be the following:
  - Old inteface `pbomb 30 400` should simulate 30 photons of 400nm. 
  - New interface `pbomb 30 LBSPECTRA[400nm]` should simulate 30 photons whose wavelength/energy are pulled from the spectrum defined under tblname `LBSPECTRA` and index `400nm`.  (BTW, from an implementation standpoint, it's probably ok to hardcode the tablename, so you can just look for the words `LBSPECTRA` in your state string, rather than coming up with some fancy thing to parse the strings and determine whether it's a number or not). 
 
Does this sound ok with you? Others such as @tannerbk should probably also see if this interface is agreeable.

Thanks again!

src/gen/src/VertexGen_PhotonBomb.cc
srig19: I updated the PR in EosSimulations to change the name, our ratdb table is still called `LBSPECTRA.ratdb` but the name is still "WLSPECTRA"

PR #251, opened 2025-03-31
Bonsai dev

anikakatt:
Edited bonsai to use the fitter interface to grab PMT hit times and charges.


tannerbk:
Thanks @anikakatt. Please remove any Eos specific tables and move them to the EosSimulations repo.

PR #252, opened 2025-04-01
Fully Deprecate WaveformAnalysis

JamesJieranShen:
Remove WaveformAnalysis from the repo entirely. This PR contains two commits. The first removes the instantiation of the WaveformAnalysis processors, which generated extraneous deprecation warnings in the log (thanks to @baldonia for reporting this). The second one completely removes it. If people think this is too aggressive for now, I can remove the second commit.


tannerbk:
LGTM.

PR #253, opened 2025-04-01
Small changes related to DAQ and waveform analysis

tannerbk:
1. Remove some unused headers in DAQ modules
2. Add estimated PE to DS::DigtPMT, to be filled by any method that attempts to estimate the number of PEs for a given PMT hit. This is separate from some method that attempts to extract the hit-times/hit-charges for each individual PE, which is handled in the WaveformAnalysisResults. Happy to discuss alternative implementations.
3. Change "getNhits" to "getNPEs" so that function is more appropriately named. 
4. Add some additional waveform analysis variables to the ntuple.


src/io/src/OutNtupleProc.cc
```diff
@@ -505,17 +511,20 @@ Processor::Result OutNtupleProc::DSEvent(DS::Root *ds) {
         digitPMTID.push_back(digitpmt->GetID());
         digitTime.push_back(digitpmt->GetDigitizedTime());
         digitCharge.push_back(digitpmt->GetDigitizedCharge());
+        digitNCrossings.push_back(digitpmt->GetNCrossings());
+        digitTimeOverThreshold.push_back(digitpmt->GetTimeOverThreshold());
+        digitPredictedPE.push_back(digitpmt->GetEstimatedPE());
         if (digitpmt->GetNCrossings() > 0) {
           digitNhits++;
         }
-        digitNCrossings.push_back(digitpmt->GetNCrossings());
```
JamesJieranShen, line 511: I don't think digitNCrossings should be removed here?

src/ds/include/RAT/DS/DigitPMT.hh
```diff
@@ -80,6 +80,10 @@ class DigitPMT : public TObject {
   virtual void SetTimeOffset(Double_t _time_offset) { this->time_offset = _time_offset; }
   virtual Double_t GetTimeOffset() { return time_offset; }
 
+  /** Estimated number of photoelectrons, from likelihood algorithm */
+  virtual void SetEstimatedPE(Int_t _fNPE) { this->fNPE = _fNPE; }
```
JamesJieranShen, line 84: The naming here kind of breaks convension -- everything in a digitPMT is supposedly `Estimated` -- so I would suggest just name this `SetNPEs`. Also, it should be plural.

src/io/include/RAT/OutNtupleProc.hh
```diff
@@ -156,7 +156,10 @@ class OutNtupleProc : public Processor {
   std::vector<double> digitPeak;
   std::vector<double> digitTime;
   std::vector<double> digitCharge;
+  std::vector<double> digitTimeOverThreshold;
+  std::vector<double> digitVoltageOverThreshold;
   std::vector<double> digitLocalTriggerTime;
+  std::vector<int> digitPredictedPE;
```
JamesJieranShen, line 162: This should certainly be the same as whatever name we using in `digitPMT` -- either `estimatedPEs` or just `NPEs`

JamesJieranShen (changes requested):
Looks good to me in general! Couple small tweaks.

src/io/src/OutNtupleProc.cc
tannerbk: This is moved up, I can put it back below the nhits counter if it is preferred.

tannerbk:
@JamesJieranShen should be ready for review.

JamesJieranShen (approved):

PR #254, opened 2025-04-01
Fix probability check in PMTOpticalModel

### LINKED ISSUE #242: efficiency greater than 1
We are having the issue with efficiency greater than 1 error:
https://github.com/rat-pac/ratpac-two/blob/ab24a151732a84199bc5665bc379714273de4130/src/physics/src/GLG4PMTOpticalModel.cc#L392

The efficiency peak value in our model is 0.64. The thin_factor is always 1. It gives the efficiency > 1 error.
Can you have an option to comment that line out?

llebanowski:
Addresses Issue #242 [@gyang9].  This issue was independently observed by @ZachL00 and this PR was found to resolve it.  

Issues: Tests of bounds on collection_eff and A (absorption probability) were inside a debug conditional, meaning they did not run in general.  If out of bounds, both collection_eff and A were set to 0.5, which is entirely arbitrary.  
Without changing the core behavior, this PR:

1.   Tests bounds outside of debug mode.
2.   Corrects one of the bounds to use the p.e. probability (A*collefction_eff) instead of just collection_eff, which could have a value greater than 1 without being incorrect.
3.   Assigns reasonable values to out of bounds parameters.

fixes #242

### DIFF
--- a/src/physics/src/GLG4PMTOpticalModel.cc
+++ b/src/physics/src/GLG4PMTOpticalModel.cc
@@ -377,15 +377,13 @@ void GLG4PMTOpticalModel::DoIt(const G4FastTrack &fastTrack, G4FastStep &fastSte
     An = 1.0 - (fT_n + fR_n);           // The absorption at normal incidence
     collection_eff = _efficiency / An;  // net QE = _efficiency for normal inc.
 
-#ifdef G4DEBUG
     if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || A * collection_eff > 1.0) {
+#ifdef G4DEBUG
       RAT::debug << "GLG4PMTOpticalModel::DoIt(): Strange coefficients!" << newline;
       RAT::debug << "T, R, A, An, weight: " << T << " " << R << " " << A << " " << An << " " << weight << newline;
       RAT::debug << "collection eff, std QE: " << collection_eff << " " << _efficiency << newline;
       RAT::debug << "=========================================================" << newline;
-    }
 #endif
-    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || A * collection_eff > 1.0) {
       collection_eff = _efficiency;
       if (A <= 0.0) {
         A = 0.0;
--- a/src/physics/src/GLG4PMTOpticalModel.cc
+++ b/src/physics/src/GLG4PMTOpticalModel.cc
@@ -378,18 +378,25 @@ void GLG4PMTOpticalModel::DoIt(const G4FastTrack &fastTrack, G4FastStep &fastSte
     collection_eff = _efficiency / An;  // net QE = _efficiency for normal inc.
 
 #ifdef G4DEBUG
-    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || collection_eff > 1.0) {
+    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || A * collection_eff > 1.0) {
       RAT::debug << "GLG4PMTOpticalModel::DoIt(): Strange coefficients!" << newline;
       RAT::debug << "T, R, A, An, weight: " << T << " " << R << " " << A << " " << An << " " << weight << newline;
       RAT::debug << "collection eff, std QE: " << collection_eff << " " << _efficiency << newline;
       RAT::debug << "=========================================================" << newline;
-      A = collection_eff = 0.5;  // safe values???
     }
 #endif
+    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || A * collection_eff > 1.0) {
+      A = 1.0;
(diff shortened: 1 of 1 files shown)


src/physics/src/GLG4PMTOpticalModel.cc
```diff
@@ -378,18 +378,26 @@ void GLG4PMTOpticalModel::DoIt(const G4FastTrack &fastTrack, G4FastStep &fastSte
...
-    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || collection_eff > 1.0) {
+    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || A * collection_eff > 1.0) {
       RAT::debug << "GLG4PMTOpticalModel::DoIt(): Strange coefficients!" << newline;
       RAT::debug << "T, R, A, An, weight: " << T << " " << R << " " << A << " " << An << " " << weight << newline;
       RAT::debug << "collection eff, std QE: " << collection_eff << " " << _efficiency << newline;
       RAT::debug << "=========================================================" << newline;
-      A = collection_eff = 0.5;  // safe values???
     }
 #endif
+    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || A * collection_eff > 1.0) {
```
JamesJieranShen, line 388: Can we combine the two if blocks here? The condition is the same, so the preprocessor directive can just live inside the if block.

src/physics/src/GLG4PMTOpticalModel.cc
```diff
@@ -378,18 +378,26 @@ void GLG4PMTOpticalModel::DoIt(const G4FastTrack &fastTrack, G4FastStep &fastSte
...
+      collection_eff = _efficiency;
+      if (A <= 0.0) {
+        A = 0.0;
+      } else {
```
JamesJieranShen, line 392: Do we actually want this to be `else if (A > 1.0)`? Presumably if the collection efficiency is too large and we had to scale it to `_efficiency`, A would not need to be scaled.

src/physics/src/GLG4PMTOpticalModel.cc
```diff
@@ -378,18 +378,26 @@ void GLG4PMTOpticalModel::DoIt(const G4FastTrack &fastTrack, G4FastStep &fastSte
...
+        A = 1.0;
+      }
+    }
 
     collection_eff *= RAT::PhotonThinning::GetFactor();
-    if (collection_eff > 1.0) {
-      RAT::Log::Die(dformat("PMT collection efficiency of %f is >1.0! Is thin_factor too big?", collection_eff));
+    if (A * collection_eff > 1.0) {
```
JamesJieranShen, line 398: I understand this was  the old behavior, but perhaps we shouldn't die here, but rather just shout out a really loud warning? I'm not sure.

src/physics/src/GLG4PMTOpticalModel.cc
llebanowski: The _efficiency should almost certainly not be at fault, rather the An, which is effectively set to 1 here by setting collection_eff to _efficiency.  Therefore, I suggest also setting A to 1.

src/physics/src/GLG4PMTOpticalModel.cc
llebanowski: I think it should die here because the physics will be incorrect if it proceeds.

JamesJieranShen (approved):
Thanks for the response, Logan, I'm happy with this change now given your reasons.

src/physics/src/GLG4PMTOpticalModel.cc
```diff
@@ -377,19 +377,25 @@ void GLG4PMTOpticalModel::DoIt(const G4FastTrack &fastTrack, G4FastStep &fastSte
...
 #ifdef G4DEBUG
-    if (A < 0.0 || A > 1.0 || collection_eff < 0.0 || collection_eff > 1.0) {
       RAT::debug << "GLG4PMTOpticalModel::DoIt(): Strange coefficients!" << newline;
       RAT::debug << "T, R, A, An, weight: " << T << " " << R << " " << A << " " << An << " " << weight << newline;
       RAT::debug << "collection eff, std QE: " << collection_eff << " " << _efficiency << newline;
       RAT::debug << "=========================================================" << newline;
-      A = collection_eff = 0.5;  // safe values???
-    }
 #endif
+      collection_eff = _efficiency;
```
tannerbk, line 387: Why is the collection efficiency reset to the quantum efficiency here, for example in the cases where A < 0 or A > 1

llebanowski: Implemented more complete checks and resets.  I believe these checks will not fail in general because the previous spurious failure was due to the inappropriate check of collection_eff when it should have been checking A*collection_eff.

PR #255, opened 2025-04-09
added fitted charge calibration

rl23:


syjeon162:
Can I ask about the motivation for the changes in `WaveformAnalysisLognormal.cc`? My understanding is that the charge scale is already reflected in the simulated pulses, and since we're fitting these pulses, we should not need to multiply the lognormal fitted charge with the charge scale again. Perhaps this is why the charge scales have been applied twice.

tannerbk:
Right now the PMT pulses are being scaled by the values in the "charge_scale" table. I think we just want to make it such that we can load a different table by specifying an index. That feature already exists, but I think we want separate indices for the different tables here:
```
https://github.com/rat-pac/ratpac-two/blob/959e597f5875f1c832b76a355c64a86f84ed8d55/src/ds/include/RAT/DS/ChannelStatus.hh#L56
```

src/ds/include/RAT/DS/ChannelStatus.hh
```diff
@@ -58,8 +58,10 @@ class ChannelStatus : public TObject {
     default_offset = lCableOffset->GetD("default_value");
     DBLinkPtr lChannelOnline = DB::Get()->GetLink("channel_online", index);
     default_is_online = lChannelOnline->GetD("default_value");
-    DBLinkPtr lChargeScale = DB::Get()->GetLink("charge_scale", index);
-    default_charge_scale = lChannelOnline->GetD("default_value");
+    DBLinkPtr lChargeScaleIndex = DB::Get()->GetLink("charge_scale", "selected_charge_scale");
```
tannerbk, line 61: I think we need to update the CHANNEL_STATUS_DEFAULTS.ratdb to add this.

src/io/include/RAT/OutNtupleProc.hh
```diff
@@ -185,7 +185,7 @@ class OutNtupleProc : public Processor {
   std::vector<std::vector<double>> trackTime;
   std::vector<std::vector<int>> trackProcess;
   std::vector<std::vector<int>> trackVolume;
-
+
```
tannerbk, line 188: Can you remove this change?

tannerbk:
LGTM. I can merge the conflicts.

tannerbk:
Merging #261 with the conflicts resolved.

PR #257, opened 2025-04-16
Waveform fitter FoMs in NtupleProc

syjeon162:
Adds branches for waveform fitter figures of merit to outntuple.
The FoMs to be included in outntuple needs to be specified in `IO.ratdb`.


JamesJieranShen (approved):
LGTM. Thanks!

PR #263, opened 2025-04-30
Allow channel status indexing to be different for every column

JamesJieranShen:
Also featuring slight channel status refactoring to make people's lives easier when they add a new status.

### DIFF
--- a/ratdb/CHANNEL_STATUS_DEFAULTS.ratdb
+++ b/ratdb/CHANNEL_STATUS_DEFAULTS.ratdb
@@ -1,4 +1,12 @@
-// This table includes all default values set for the ChannelStatus class. Custom values should be set per-experiment.
+// Default channel_status selection table. This table controls which channel_selection
+{
+  "name": "channel_status_selection",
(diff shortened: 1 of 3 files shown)


ratdb/CHANNEL_STATUS_DEFAULTS.ratdb
```diff
@@ -1,4 +1,12 @@
-// This table includes all default values set for the ChannelStatus class. Custom values should be set per-experiment.
+// Default channel_status selection table. This table controls which channel_selection
+{
+  "name": "channel_status_selection",
+  "index": "",
+  "run_range": [0, 0],
+  "default": "",
+}
```
tannerbk, line 7: Can you provide an example here, commented out maybe, of how to use this table?

JamesJieranShen: done.

tannerbk:
LGTM. I tested for the charge scales and cable delays, and worked well.

PR #265, opened 2025-05-10
Fix for quad points at center of detector

smnaugle:
Quad points with uninvertable matrices were not being skipped, leading to many points being placed at (0,0,0,0) in the quad cloud. Code now checks if the matrix is invertible to skip these points.


smnaugle:
I'm not totally sure why so many matrices are not invertible. After some debugging, I think it happens when 3 or 4 PMTs in a row are selected to make a quad point, and since the distances between each of the PMTs is the same in one of the x,y directions the `M` matrix ends up being uninvertible because it has linearly dependent rows. But that's just a guess right now.

JamesJieranShen (approved):

PR #269, opened 2025-05-13
Some documentation changes

tannerbk:
- Add a README to to the docs folder to describe how to setup the documentation locally
- Remove a significant amount of out-dated documentation related to the processors
- Some small labeling additions to a couple of files, to reference from my processor documentation
- Significant edits to the processor documents, which is not yet finished because it the reconstruction parts need significant updates, to be added in the future. 
- Add new, empty sections for documentation that needs to be filled in.


gmwendel:
### Grammar

**doc/README.md:** 
“install the requirement” should be “install the requirements.”

**doc/users_guide/daq.rst:** 
“the simples triggering scheme” should be “the simplest triggering scheme.”
“we coudl then check” → “we could then check.”
“Based on whether its enabled” → “Based on whether it’s enabled.”
“events seprated in time” → “events separated in time.”
“at there is a least one hit” → “at least one hit.”

**doc/users_guide/processors.rst:**
“they … receives events” → “they … receive events.”
“data aquisition” → “data acquisition.”

**doc/users_guide/reconstruction.rst:** 
item numbering under “Path Fitter” jumps from 2. to “2)”; should continue as 3.

### Documentation Logic

**daq.rst**

Double check the parameters listed for ForcedTriggerProc.  

The processor seems to pull its configuration from a RATDB entry rather than using `/rat/procset`. At the start of a run it reads digitizer_name and digitize directly from the database:

```
void ForcedTriggerProc::BeginOfRun(DS::Run *run) {
  // Trigger Specifications
  ldaq = DB::Get()->GetLink("DAQ", "ForcedTrigger");
  fEventCounter = 0;
  fDigitizerType = ldaq->GetS("digitizer_name");
  fDigitize      = ldaq->GetZ("digitize");
  ...
}
```

tannerbk:
Thanks, good catches!

PR #270, opened 2025-05-14
Add a word for event cleaning

JamesJieranShen:
The mask is set to zero and currently does nothing except being written to the ntuples. It is up to experiments to implement the cleaning logic.


tannerbk:
Looks good. In principle it needs documentation, but I think we will add that in the upcoming documentation campaign.

PR #274, opened 2025-05-16
Clean up a bunch of outdated processor code

tannerbk:
Not ready to merge, this requires some discussion. Generally, I think we should clean-up mostly unused or highly specialized code that is not particularly appropriate for ratpac-two.

Here is my logic:

TrueDAQ -- Remove because it is HIGHLY specific to a single scenario (IBDs) and has hard-coded values all over the place. This is the type of thing a specific experiment could include, but its not appropriate for ratpac-two.

LessSimpleDAQs -- Outdated versions of splitevdaq, which is validated and more configurable. 

MiniCleanCharge/MultiCharge -- Ditto what I said for TrueDAQ. Too specific and too many hard-coded values to be useful.


MarcFBergevin:
I think this is fine, mostly. 

I am conflicted about removing LessSimpleDAQ. It handles well Geant4 events that leads to multiple triggers (IBD events, fast neutron events, Li-9 events). I have not thoroughly tested this with SplitEVDAQ. It is also convenient if you would just want TQ out of an event and do not plan to do waveform processing.

tannerbk:
Thanks for looking at this Marc. Splitevdaq should work perfectly well for multiple triggered events, and the waveform processing is entirely optional (specified in the DAQ parameters). It has been tested fairly robustly for neutron captures in Eos, but there are probably other use-cases we may want to look at before deleting LessSimpleDAQ, if you have validated that DAQ in WATCHMAN (or somewhere else).

tannerbk:
Now this also includes a slight reorganization for the lognormal fitting waveform analysis code to format it the same way as the gaussian fitting.

tannerbk:
Its my opinion that "LessSimpleDAQ" should be migrated to a private repo, unless someone wants to clean up the code and document it.

PR #276, opened 2025-05-17
Add a hit cleaning word to DigitPMT.

smnaugle:
Uses a 64 bit unsigned int to keep track of hit cleaning for DigitPMT waveforms.

### DIFF
--- a/src/ds/include/RAT/DS/DigitPMT.hh
+++ b/src/ds/include/RAT/DS/DigitPMT.hh
@@ -110,13 +110,27 @@ class DigitPMT : public TObject {
    * @param bit the literal bit position
    * @param val the value to write.
    */
-  virtual void SetHitCleaningMask(uint8_t bit, bool val) {
-    u_int32_t mask = val << bit;
-    hit_cleaning_mask = hit_cleaning_mask | mask;
+  virtual void SetHitCleaningBit(int bit, bool val = true) {
+    if (bit > (sizeof(hit_cleaning_mask) * 8) - 1) {  // Bits to bytes, then 0-indexing
+      warn << "Tried to set bit out of hit cleaning bitmask range, ignoring." << newline;
+      return;
+    }
+    uint32_t mask = (1 << bit);
+    if (val) {
+      hit_cleaning_mask = hit_cleaning_mask | mask;
+    } else {
+      hit_cleaning_mask = hit_cleaning_mask & (~mask);
+    }
   }
 
   /** Retrieve hit cleaning mask */
-  virtual uint32_t GetHitCleaningMask() { return hit_cleaning_mask; }
+  virtual bool CheckHitCleaningBit(uint8_t bit) {
+    uint64_t mask = 1 << bit;
+    return (hit_cleaning_mask & mask);
+  }
+
+  /** Retrieve hit cleaning mask */
+  virtual uint64_t GetHitCleaningMask() { return hit_cleaning_mask; }
 
   ClassDef(DigitPMT, 6);
 
@@ -137,7 +151,7 @@ class DigitPMT : public TObject {
   Double_t local_trigger_time = -9999;
   Double_t time_offset = 0;
   std::map<std::string, WaveformAnalysisResult> fit_results;
-  uint32_t hit_cleaning_mask = 0x0;
+  uint64_t hit_cleaning_mask = 0x0;
 };
 
 }  // namespace DS
--- a/src/ds/include/RAT/DS/DigitPMT.hh
+++ b/src/ds/include/RAT/DS/DigitPMT.hh
@@ -105,6 +105,19 @@ class DigitPMT : public TObject {
     return fitter_names;
   }
 
+  /**
+   * Set a bit in the hit cleaning mask.
+   * @param bit the literal bit position
+   * @param val the value to write.
+   */
+  virtual void SetHitCleaningMask(uint8_t bit, bool val) {
+    u_int32_t mask = val << bit;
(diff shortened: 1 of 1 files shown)


src/ds/include/RAT/DS/DigitPMT.hh
```diff
@@ -105,7 +105,38 @@ class DigitPMT : public TObject {
...
   }
 
-  ClassDef(DigitPMT, 6);
+  /**
+   * Set a bit in the hit cleaning mask.
+   * @param bit the literal bit position
+   * @param val the value to write.
+   */
+  virtual void SetHitCleaningBit(uint bit, bool val = true) {
+    if (bit > (sizeof(hit_cleaning_mask) * 8) - 1) {  // Bits to bytes, then 0-indexing
```
JamesJieranShen, line 114: Using `std::numerical_limits<uint64_t>::digits` here will probably make it more readable.

src/ds/include/RAT/DS/DigitPMT.hh
```diff
@@ -105,7 +105,38 @@ class DigitPMT : public TObject {
...
+      warn << "Tried to set bit out of hit cleaning bitmask range, ignoring." << newline;
+      return;
+    }
+    uint32_t mask = (1 << bit);
```
JamesJieranShen, line 118: should be `uint64_t`.

src/ds/include/RAT/DS/DigitPMT.hh
```diff
@@ -105,7 +105,38 @@ class DigitPMT : public TObject {
...
+    if (val) {
```
JamesJieranShen, line 119: This if statement is not super readable to me, could we change it to: `hit_cleaning_mask = (hit_cleaning_mask & ~(1 << n)) | (value  << n);`?

src/ds/include/RAT/DS/DigitPMT.hh
```diff
@@ -105,7 +105,38 @@ class DigitPMT : public TObject {
     return fitter_names;
   }
 
-  ClassDef(DigitPMT, 6);
+  /**
+   * Set a bit in the hit cleaning mask.
+   * @param bit the literal bit position
+   * @param val the value to write.
+   */
+  virtual void SetHitCleaningBit(uint bit, bool val = true) {
```
JamesJieranShen, line 113: Can we rename `bit` to `bit_position`, it is not very clear with the current name what is meant here. Ditto for the other functions.

JamesJieranShen (commented):
Thanks @smnaugle ! Please see review comments. Also, like discussed during the meeting today, I think we would at least like to have `nhit_cleaned` or something like this in the EV class / ntuples.

Another thing to note is that I think we need equivalent logic in the `DS::PMT` class. But that would be effectively the same code, just copied to a different class. It also doesn't have to be part of this PR.

src/ds/include/RAT/DS/DigitPMT.hh
```diff
@@ -105,7 +105,38 @@ class DigitPMT : public TObject {
...
+    uint32_t mask = (1 << bit);
+    if (val) {
+      hit_cleaning_mask = hit_cleaning_mask | mask;
+    } else {
+      hit_cleaning_mask = hit_cleaning_mask & (~mask);
+    }
+  }
+
+  /** Check a bit of the hit cleaning mask */
+  virtual bool CheckHitCleaningBit(uint bit) {
```
JamesJieranShen, line 127: Sorry, forgot to mention that this should be named with `Get`, not `Check`.

src/ds/include/RAT/DS/EV.hh
```diff
@@ -80,8 +80,27 @@ class EV : public TObject {
   virtual Int_t GetDigitPMTCount() const { return digitpmt.size(); }
   virtual void PruneDigitPMT() { digitpmt.clear(); }
 
+  const std::vector<Int_t> GetAllCleanedDigitPMTIDs() const {
+    std::vector<Int_t> result;
+    for (auto const &kv : digitpmt) {
+      auto hit_cleaning_mask = kv.second.GetHitCleaningMask();
```
JamesJieranShen, line 86: Can we not use `auto` on this line? IMO  `auto` should really only be used when the inferred type is immediately obvious.

src/ds/include/RAT/DS/EV.hh
```diff
@@ -80,8 +80,27 @@ class EV : public TObject {
...
+        result.push_back(kv.first);
+      }
+    }
+    return result;
+  }
+
+  // TODO: Implemetation for PMT class
+  const std::vector<Int_t> GetAllCleanedPMTIDs() const {
+    std::vector<Int_t> result;
+    return result;
```
JamesJieranShen, line 97: Let's just have this `Die` for now -- I think it would be quite confusing if the users just see an empty array..

JamesJieranShen (commented):
Thanks again @smnaugle -- a couple more comments.

JamesJieranShen:
> Thanks for the comments @JamesJieranShen, I think this commit resolves most of them. I didn't change the bit_position size check since my implementation uses the actual size of the `hit_cleaning_mask`, so if we ever decide to change the type of that variable we don't have to worry about changing the checks.

As  for this -- Can we just `typedef` the type out and still use numerical limits? Doing these very c-oriented things gets me a bit scared. Using `numerical_limits`, is self-documenting on the other hand.

smnaugle:
OK I think this addresses all the comments. Does this look good @JamesJieranShen ?

PR #277, opened 2025-05-19
Fix Nhit in fitdirectioncenter

llebanowski:


JamesJieranShen:
Thanks Logan. There's also the seeding feature added in #271 that we should incorporate in this fitter. Would you like us to merge this now and worry about that later or would you be Ok with waiting til that PR is merged and combine then have this PR include the seeding changes?

llebanowski:
Hi James, thanks for that note.
I think merging this pull request first would be most helpful and cleanest.

macros/examples/fitdirectioncenter.mac
```diff
@@ -40,6 +41,7 @@
 /rat/procset time_resid_up  0.7                           # upper cut on time residuals using direct cut [ns]
 /rat/procset time_resid_low 0.06                          # lower cut on time residuals using direct cut [ns]
 /rat/procset verbose 1                                    # save additional parameter in data structure
+/rat/procset pmt_type 0                                   # type of pmt to be used in reconstruction
 /rat/procset pmt_type 1                                   # type of pmt to be used in reconstruction
```
tannerbk, line 45: There are only type = 0 PMTs in the Validation geo, so we may want to remove the others since it could cause confusion.

llebanowski: Multiple lines for `pmt_type` are shown to illustrate how to select multiple PMT types.  
Your note is correct and so for the validation geo, we don't need to write any of these lines.  
I think it is helpful to illustrate this somewhere and I am glad to move these lines if a more preferred location is identified, or to modify the comments to reduce confusion.

tannerbk: You can leave them here and just add a comment.

PR #283, opened 2025-05-21
Documentation update for RATDB and install procedure.

JamesJieranShen:
first round of doc updates, with many more to come:)


JamesJieranShen:
> Minor changes requested:
> 
> Typos: begining → beginning palce → place crerates → creates contaienr → container witha → with a ratpc-two → ratpac-two
> 
> Grammar: ones that start with // and surrounded by /* and */ → ones that start with // or are surrounded by / and / it allows the key → it allows the keys The plane of which → The plane in which There's also tables that utilizes → There are also tables that use and indicates → and it indicates It's use → Its use RATDB also have → RATDB also has one need to set → one needs to set

Resolved. Thanks for the spell checks. Please use the review feature next time:)

PR #286, opened 2025-05-23
Addition of nested tube hits option in IO.ratdb

ZachL00:
Addition of nested tube hits option in IO.ratdb to avoid failure of try/catch statement in src/io/src/OutNtupleProc.cc.


tannerbk:
Good catch, LGTM.

PR #289, opened 2025-05-27
Add python processor example, remove charge balance classifier

tannerbk:


JamesJieranShen:
Thanks. I finally took a look at the python processor code and it's .. surprisingly simple? I do think we should rename the methods to conform with the C++ code -- there should be `dsevent`, `event` (as they exist right now). `finish` should probably be renamed `endOfRun`, and we should add a `beginOfRun`. They should also be moved out of the destructor and into the corresponding C++ functions. Let me know if you would like to include these changes in this PR. If not, I'm happy to work on it.

tannerbk:
I believe you can run the Hist processor by doing:
`/rat/proc python
/rat/procset class "ratproc.Hist('GetEVCount()', 3, 0, 3, interval=1)"`
We can extend the documentation soon, particularly if James changes some of the naming as he suggested.

PR #293, opened 2025-06-04
Water world + PMT encapsulation for Visualization purposes

MarcFBergevin:
Adding a world filled with water and nothing else for RATPAC2 summer school. 

Also added is a visualization macro to take this world and add two PMTs with encapsulation on. Encapsulation are currently being implemented in PR 256.


MarcFBergevin:
@tannerbk  This could technically be deleted, but I do like the idea of having an empty world that can just be loaded. Maybe the name is not the best.

PR #295, opened 2025-06-13
Add michel electron energy spectrum to the spectrum vertex generator

hbjamin:


hbjamin:
The Michel energy spectrum plotted is ` 3x**2 - 2x**3`  where `x = E_e / E_max` and `E_max = 52.8`. This can be derived by integrating equation `58.3` from the reference below over `cos(theta)`. The spectrum is normalized so that the integral from `x = 0` to `x = E_max` is equal to `1`. 

https://pdg.lbl.gov/2020/reviews/rpp2020-rev-muon-decay-params.pdf?utm_source=chatgpt.com

tannerbk:
Thanks Ben. Can you please add a comment in the ratdb table specifying how the energy spectrum was calculated (similar to your github comment)?

tannerbk:
Sorry, this needs to be documented in the user guide as well, as part of the generators.

Issue #157, opened 2024-08-08
Fully deprecate valid_begin and valid_end from ratdb


JamesJieranShen:
many ratdb tables still use `valid_begin` and `valid_end`, instead of the replacement parameters `run_range`. There's no longer any documentation or code that explains how to use these old parameters, so they should be removed from all DB tables. For reference, sno+ did this almost  a decade ago;)

PR #296, opened 2025-06-13
Deprecate valid_begin, valid_end from ratdb

syjeon162:
Addresses https://github.com/rat-pac/ratpac-two/issues/157.
* Replaces `valid_begin` and `valid_end` with `run_range` in all ratdb & geo files.
* Replaces relevant code in `DB`. Any table that uses the deprecated syntax will get a warning message.
* ...and also adds quotation marks around all json keys...


tannerbk:
LGTM.

PR #299, opened 2025-06-19
Fix bug in WaveformUtil::CalculateTimeCFD

syjeon162:
Commit 0ae95141812443d22d54737a302090684b0c4c9f :

> Previously, line 69-74...
> ```c++
>    // Reached the begining of the waveform
>    // returned an invalid value
>    if (i == back_window) {
>      thresholdCrossing = INVALID;  // Invalid value for bad waveforms
>      break;
>    }
>``` 
> ...was not doing anything, since we loop until `i > back_window`.
> Hence when we don't cross threshold, `GetThresholdCrossingBeforePeak` returned `thresholdCrossing = 0`.
> This causes `CalculateTimeCFD` to return an unrealistic number -- often some small negative value that is outside the trigger window.
>
> Instead, initiate `thresholdCrossing` as `INVALID`, and let it return `INVALID` if threshold is not crossed.

Commit 42842ab3ab09545c2bfb9bf051a7f8e6e4aba629 : 

> Error catching, although this shouldn't happen -- we should never interpolate "backwards".

Commit e30ca73e61581eea183e03363c9ef13b3d7e311b : 

> If the peak is not above threshold, return `INVALID`.
> Previously it returned `peakSample` since it meets the `break` condition immediately.

### DIFF
--- a/ratdb/DIGITIZER_ANALYSIS.ratdb
+++ b/ratdb/DIGITIZER_ANALYSIS.ratdb
@@ -7,7 +7,7 @@ valid_end: [0,0],
 pedestal_window_low: 0,
 pedestal_window_high: 10,
 constant_fraction: 0.40,
-// Number of samples to look back after the peak
+// Time to look back after the peak in ns
 lookback: 20.0,
 // Integration window, in samples
 integration_window_low: 4,
--- a/src/ds/include/RAT/DS/DigitPMT.hh
+++ b/src/ds/include/RAT/DS/DigitPMT.hh
@@ -13,6 +13,7 @@
 
 #include <RAT/DS/WaveformAnalysisResult.hh>
 #include <RAT/Log.hh>
+#include <RAT/WaveformUtil.hh>
 #include <limits>
 
 namespace RAT {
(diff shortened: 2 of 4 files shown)


syjeon162:
Some plots of what this actually changes:
Screenshot 2025-06-18 at 9 40 46 PM

syjeon162:
Yes, `digitTime` should only be `INVALID` if threshold crossing was invalid.
There are two main cases:
(1) Waveforms are actually too *large* and doesn't cross threshold within the `lookback` range (stays above threshold). Perhaps this is important to note and individual experiments might want to consider increasing this lookback range.
(2) The waveform is below threshold. This usually happens for logic signals that are given some absolute threshold instead of a constant fraction.

And yes, the plots do include the `time_offset` subtraction -- sorry about the confusion.

I just pushed a commit to not subtract `time_offset`s for `INVALID` values in `DigitPMT` and also for waveform analyses. Please take a look! Also think it would be good if someone else could test this and make sure it's not introducing any new unexpected behavior. Thanks.

rpitelka:
Everything looks good, but I'd like to understand where the `INVALID` `digitTimes` are coming from. I tried adjusting the `constant_fraction` and `lookback` parameters in `DIGITZER_ANALYSIS.ratdb` but still saw bad values, and couldn't find any correlations with the amplitude or mc hit time. Maybe bad pedestals?

syjeon162:
I took a quick look at the actual waveforms that are marked as invalid and it seems to be mostly three things:
(1) There's a double peak, (2) The waveform is too early in the digitization window and the rising edge is clipped, (3) a lot of noise with a drifting baseline. Here are some examples of each.
Screenshot 2025-06-24 at 5 50 05 PM

rpitelka:
Ah great, thanks! The noise case seems like an appropriate `INVALID` return. Maybe we should consider extending the lookback window to catch the double peak case? And for the early in-window case, since `digitTime` is used to seed the Lognormal and Gaussian fits, it might be helpful to return 0 instead if the beginning of the waveform is reached. Let me know your thoughts on this.

syjeon162:
I think that sounds reasonable! I handled the early in-window cases in `CalculateTimeCFD` because we also need to skip interpolation, but let me know what you think. I've tested this on the dataset I've been looking at and seems to work well.

The `lookback` range in the default ratpac-two ratdb is already quite a bit longer than what I've been working with (10ns) and actually would have caught most of the double peak cases I'm seeing -- perhaps it should be up to individual experiments to change that.

rpitelka:
Tested and looks good to me! Ready for final review @JamesJieranShen.

JamesJieranShen (approved):
LGTM!

PR #301, opened 2025-06-22
Add Begin/EndOfRun logic to InROOTProducer

JamesJieranShen:
In the long term inroot should properly support multi-file read. Right now it doesn't, but this PR makes it at least minimally usable.

It would be nice if this can go in before the summer school session on this topic:)

### DIFF
--- a/src/io/src/InROOTProducer.cc
+++ b/src/io/src/InROOTProducer.cc
@@ -102,14 +102,26 @@ bool InROOTProducer::ReadEvents(G4String filename) {
 
   // Read
   Int_t num_events = tree.GetEntries();
+  DS::Run *run;
   for (Int_t i = 0; i < num_events && !SignalHandler::IsTermRequested(); i++) {
     tree.GetEntry(i);
     // force the run entry to be loaded into memory so that it
     // can be written later.
     // If no runTree to read from, this will return 0, but we don't care.
+    if (i == 0) {
+      run = DS::RunStore::GetRun(branchDS);
+      mainBlock->BeginOfRun(run);
+    } else if (branchDS->GetRunID() != run->GetID()) {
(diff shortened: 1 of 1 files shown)


src/io/src/InROOTProducer.cc
```diff
@@ -102,15 +102,27 @@ bool InROOTProducer::ReadEvents(G4String filename) {
...
     // If no runTree to read from, this will return 0, but we don't care.
+    if (i == 0) {
+      run = DS::RunStore::GetRun(branchDS);
+      mainBlock->BeginOfRun(run);
+    } else if (branchDS->GetRunID() != run->GetID()) {
+      mainBlock->EndOfRun(run);
+      run = DS::RunStore::GetRun(branchDS);
+      mainBlock->BeginOfRun(run);
+    }
     DS::RunStore::GetRun(branchDS);
```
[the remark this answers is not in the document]
JamesJieranShen: The rootfiles can be provided as a TChain, so there's no guarantee that the files all come from the same run. Is that what you mean?

src/io/src/InROOTProducer.cc
```diff
@@ -102,15 +102,27 @@ bool InROOTProducer::ReadEvents(G4String filename) {
...
 
     mainBlock->DSEvent(branchDS);
   }
+  if (run) {
```
gmwendel, line 123: If run is never initialized e.g. num_events is zero, the loop does not execute, leaving run uninitialized. The final if (run) then accesses an indeterminate pointer, and might lead to some strange behavior.

JamesJieranShen: per C++ standard a pointer is set to zero upon creation, thus the if statement evaluates to false.

PR #302, opened 2025-06-23
Laserball generator

srig19:
Laserball vertex generator, takes in measured spectra loaded into `WLSPECTRA.ratdb` and generates photons with wavelength according the simulation. Verification slides: https://nino.lbl.gov/eos/DocDB/cgi/ShowDocument?docid=447


JamesJieranShen (commented):
Thanks @srig19 ! See review for code-specific comments. I have not tested this so will not comment on the physics performance.

JamesJieranShen (approved):
LGTM -- thanks for all the edits!

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -111,15 +150,21 @@ void VertexGen_PhotonBomb::SetState(G4String newValues) {
 
     if (fRndmEnergy) delete fRndmEnergy;
     fRndmEnergy = new CLHEP::RandGeneral(energyResample, nbins);
-  } else
+  } else {
     fEnergy = CLHEP::hbarc * CLHEP::twopi / (wavelength * CLHEP::nm);
```
tannerbk, line 154: `wavelength` is a string here, I don't think this will compile.

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -68,65 +85,135 @@ void VertexGen_PhotonBomb::SetState(G4String newValues) {
...
+  is >> num >> wavelengthString;
+  double exp = 0.0;
+
+  if (fDist) {
+    fWavelengthIndex = wavelengthString;
+
+    try {
+      spectradb = DB::Get()->GetLink("WLSPECTRUM", fWavelengthIndex);
+    } catch (DBNotFoundError &e) {
+      Log::Die("VertexGen_PhotonBomb: (Using Distribution) Wavelength simulated does not have measured spectrum.");
```
JamesJieranShen: This message is no longer descriptive -- it should say `Requested photon bomb distribution is not found.` or something to that effect.

JamesJieranShen (commented):
Thank you @srig19 ! Please see additional comments. Additionally, can you add documentation on this change as well?

tannerbk (commented):
I think this should be ready other than one small comment.

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -33,10 +38,14 @@ void VertexGen_PhotonBomb::GeneratePrimaryVertex(G4Event *event, G4ThreeVector &
 
     // Use fixed energy unless spectrum was provided
     double energy;
-    if (fRndmEnergy)
+    if (fRndmEnergy) {
       energy = fMinEnergy + (fMaxEnergy - fMinEnergy) * fRndmEnergy->shoot();
-    else
+    } else if (fSpectrum) {
+      double wavelength = pickWavelength(fWavelengths, fIntensities);
```
gmwendel: This may be a relatively expensive call to make for generating photons.  It seems the integral for building the CDF vector is done from scratch for every single photon generated, which could be computationally expensive for large numbers of photons. The CDF should be calculated once and cached if possible.

gmwendel:
This PR appears to break backward compatibility by removing the ability to provide a photon spectrum by pointing to a material's scintillation properties in ratdb (likely useful for debugging).

These changes also introduce an algorithmic divergence for two generators that simulate the same physical process. Gen_LED handles spectra using RandGeneral (resampling), while PhotonBomb now uses pickWavelength (inverse CDF). Maintaining two different sampling implementations for the same physical process is inefficient and will likely lead to inconsistent/confusing simulation results.

JamesJieranShen:
> This PR appears to break backward compatibility by removing the ability to provide a photon spectrum by pointing to a material's scintillation properties in ratdb (likely useful for debugging).
> 
> These changes also introduce an algorithmic divergence for two generators that simulate the same physical process. Gen_LED handles spectra using RandGeneral (resampling), while PhotonBomb now uses pickWavelength (inverse CDF). Maintaining two different sampling implementations for the same physical process is inefficient and will likely lead to inconsistent/confusing simulation results.

Per discussion offline: 
 - I think it is fine to have the scint. spectrum feature removed -- we can just have users copy the RATDB tables. This makes the behavior of the laserball generator a lot more predictable. If this becomes a common task, we should write a new generator for it.
 - The CLHEP generator uses the same method for sampling as the manually written one. I propose that we should use `TH1` from ROOT for this as it's a higher level interface. But the divergence should not be a huge issue, hopefully.

srig19:
Ah gotcha, I thought the issue was the way I was making the CDF and using G4UniformRand() rather than using CLHEP::RandGeneral->shoot(). I've changed it so the CDF is only made once, and I'm doing some tests to make sure it's still behaving properly. But I'll update the pr soon!

src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -68,65 +86,137 @@ void VertexGen_PhotonBomb::SetState(G4String newValues) {
...
+        }
+      } catch (DBNotFoundError &e) {
+        Log::Die("VertexGen_PhotonBomb: (Using Distribution) Error with retrieving wavelength spectrum.");
+      }
+
+    } else {
+      int wavelength;
+
+      for (size_t i = 0; i < wavelengthString.size(); i++) {
+        if (!std::isdigit(static_cast<unsigned char>(wavelengthString[i])))
```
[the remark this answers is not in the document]
tannerbk: @srig19 have you tested the scintillator spectrum functionality is unchanged?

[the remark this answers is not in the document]
srig19: I didn't before, doing that now. But I noticed it doesn't work for out-of-the box ratpac-two. @bencattermole, you mentioned LiquidO uses it, do you know if it's been used recently and if any fixes had to be done to use the feature? If it needs to be fixed, I can definitely try fixing that too.

tannerbk:
Will merge today during coordination call.

PR #308, opened 2025-06-25
Add an example for opening a RAT ntuple

tannerbk:
Tried to be very explicit with the naming and comments.


JamesJieranShen (approved):
LGTM. Let me know if you want a pandas / uproot version for this as well, I'm happy to do prepare it.

PR #309, opened 2025-06-25
feat: Created 'include_opticalproperties' parameter that saves inputted optical parameters to TGraph format in output file

sgshugh:
* Adds new parameter "include_opticalproperties"
* Parameter will create new opticalProperties TDirectory in ntuple
* ntuple contains all inputted optical parameters from the ratdb tables. All are saved as TGraphss

### DIFF
--- a/doc/users_guide/outputs.rst
+++ b/doc/users_guide/outputs.rst
@@ -46,6 +46,7 @@ Parameters:
     /rat/procset include_digitizerwaveforms 1
     /rat/procset include_digitizerhits 1
     /rat/procset include_digitizerfits 
+    /rat/procset include_opticalproperties 1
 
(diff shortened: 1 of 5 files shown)


JamesJieranShen:
Thanks @sgshugh ! This is certainly a good feature to have. On the implementation side,  `RAT::DS`  classes needs to included as part of the ROOT class definitions and be written to the full ROOT files via the `outroot` processor. For your newly defined Optical class, I don't believe this is being done. In fact, I personally think that this should _not_ be included as part of the DS structure at all since it's a helper file that is only relevant to the OutNtuple processor. So perhaps this class should live in the `io` directory and being instantiated directly by the OutNtuple processor instead of being included as part of `DS::Run`. What do you think?

PR #310, opened 2025-06-25
fix: changed CMakeLists.txt to allow MacOS (arm64) compatibility

sgshugh:
* Commented set(CMAKE_OSX_ARCHITECTURES "x86_64") in CMakeLists.txt to allow arm64 architectures to link
* Added set(CMAKE_SHARED_LIBRARY_SUFFIX ".so") in src/CMakeLists.txt to force '.so' suffix on shared library output


tannerbk:
@sgshugh thanks very much for testing these changes. @JamesJieranShen I believe we've confirmed on a mac that these changes are needed?

JamesJieranShen:
> @sgshugh thanks very much for testing these changes. @JamesJieranShen I believe we've confirmed on a mac that these changes are needed?

Yes, i think so!

PR #323, opened 2025-08-05
lucyddm implementation

JamesJieranShen:
https://nino.lbl.gov/eos/DocDB/cgi/ShowDocument?docid=448

### DIFF
--- a/doc/users_guide/waveform_analysis.rst
+++ b/doc/users_guide/waveform_analysis.rst
@@ -124,6 +124,34 @@ Sinc interpolation
 
 Describe sinc interpolation.
 
+
+-------------------------
+
+Richardson-Lucy Direct De-modulation (LucyDDM)
+``````````````````
+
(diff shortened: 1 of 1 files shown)


JamesJieranShen:
uh.. looks like FFTW3 building has some issues. I will investigate this later tonight.

tannerbk:
This should be documented in the ratpac-two docs.

ratdb/IO.ratdb
```diff
@@ -24,9 +24,10 @@
 "include_mchits": true,
 "include_calib": true,
 
-"waveform_fitters": ["Lognormal", "Gaussian", "Sinc"],
+"waveform_fitters": ["Lognormal", "Gaussian", "Sinc", "LucyDDM"],
```
tannerbk, line 27: I would not add this here given that LucyDDM is quite slow.

PR #324, opened 2025-08-06
Add digitHitCleanedPMTID and digitHitCleanedNhits to output ntuple

hbjamin:
This adds `digitHitCleanedPMTID` and `digitHitCleanedNhits` to the output ntuple file, using the function Sam already created `GetAllCleanedDigtitPMTID()` in `ratpac-two/src/ds/include/RAT/DS/EV.hh` in PR #276, however this function only returns pmtids where all bits in the hit cleaning mask are 0 (pass). 

There should probably be some field we reference like the hex hit cleaning mask `hit_cleaning_mask` in `ratpac-two/ratdb/FITTER.ratdb.` Currently hit cleaning methods are enabled/disabled in `EosSimulations/ratdb/HIT_CLEANING.ratdb`, and their bit position is set in the hit cleaning processor at `EosSimulations/src/data_cleaning/hit_cleaning/src/HitCleaningProc.cc`.
 
Shoud we...
1. Move the hit cleaning processor to `ratpac-two` and add `ratpac-two/ratdb/HIT_CLEANING.ratdb`
2. Reference `ratpac-two/ratdb/FITTER.db` in `ratpac-two/src/ds/include/RAT/DS/EV.hh`
3. Do some obvious solution that eludes me

image

### DIFF
--- a/src/io/include/RAT/OutNtupleProc.hh
+++ b/src/io/include/RAT/OutNtupleProc.hh
@@ -199,6 +199,8 @@ class OutNtupleProc : public Processor {
   std::vector<int> digitReconNPEs;
   std::vector<int> digitNCrossings;
   std::vector<int> digitPMTID;
+  int digitHitCleanedNhits;
+  std::vector<int> digitHitCleanedPMTID;
   std::vector<uint64_t> digitHitCleaningMask;
   // Information from fit to the waveforms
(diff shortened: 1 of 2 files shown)


hbjamin:
I'm being stupid just going to open a new pull request

PR #331, opened 2025-09-13
Prevent the double counting of photons in Cherenkov/Scint/Reemission counters

JamesJieranShen:
This was noticed by @JohannMartyn -- thanks for the report! An assumption was previously made in `Gsim` that G4PreTrackUserAction will only be called once per track upon the track's creation. This turns out not be true -- the method can be called again when Geant4 propagates particles across volume boundaries occasionally. More importantly, the method is _guaranteed_ to be called again when a photon leaves the PMT Optical model, whenever it is transmitted or reflected. This means that photons are double counted towards the cher/scint/reemission when they are transmitted / reflected by PMTs, causing discrepancies in number of photons by RATPAC2 compared to other simulation softwares such as WCSim. 

This patch adds a check on the TrackInfo object, and ensures that a photon is only counted once per event even if PreTrackAction is called multiple times. I've tested this in the Validation geometry using `electron.mac`, the average number of cherenkov photons dropped from ~860 to around 720 for a 2.5MeV electron.

Since this number might be relevant in important optical calibrations, there may be potential implications on analysis results. Perhaps @llebanowski , @tannerbk et. al can comment on anything we need to pay attention to before merging this.

### DIFF
--- a/src/core/src/Gsim.cc
+++ b/src/core/src/Gsim.cc
@@ -303,6 +303,7 @@ void Gsim::PreUserTrackingAction(const G4Track *aTrack) {
     // grumble, grumble, C++ const keyword, grumble
     const_cast<G4Track *>(aTrack)->SetUserInformation(new TrackInfo);
   }
+  TrackInfo *trackInfo = dynamic_cast<TrackInfo *>(aTrack->GetUserInformation());
 
   // For very large, complex tracks, it is not sufficient to
   // discard the track if we do not want it.  We must prevent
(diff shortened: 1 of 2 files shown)


src/core/src/Gsim.cc
```diff
@@ -322,10 +323,12 @@ void Gsim::PreUserTrackingAction(const G4Track *aTrack) {
     fpTrackingManager->SetStoreTrajectory(false);
   }
 
-  if (aTrack->GetDefinition()->GetParticleName() == "opticalphoton") {
+  if (aTrack->GetDefinition()->GetParticleName() == "opticalphoton" &&
+      !trackInfo->preUserTrackingActionDone  // this is needed because it is _not_ guaranteed that G4 only calles this
```
tannerbk, line 327: Several typos in comment.

JamesJieranShen: thanks:) fixed

PR #332, opened 2025-09-16
rsNNLS Waveform Analysis Processor

rpitelka:
Add the rsNNLS waveform analysis processor to rat. See [here](https://nino.lbl.gov/eos/DocDB/cgi/ShowDocument?docid=466) for a description of the algorithm and some results.

### DIFF
--- a/src/daq/src/WaveformAnalysisRSNNLS.cc
+++ b/src/daq/src/WaveformAnalysisRSNNLS.cc
@@ -16,8 +16,7 @@
 namespace RAT {
 
 void WaveformAnalysisRSNNLS::Configure(const std::string& config_name) {
-  info << "WaveformAnalysisRSNNLS::Configure called with config_name: " << config_name << " (instance @ " << this << ")"
-       << newline;
+  debug << "WaveformAnalysisRSNNLS: Configure called with config_name " << config_name << newline;
   // Load analysis parameters from DIGITIZER_ANALYSIS database
   try {
     fDigit = DB::Get()->GetLink("DIGITIZER_ANALYSIS", config_name);


rpitelka:
TO DO:
 - Add numPE estimation
 - Add other threshold options (sparsity, goodness of fit)

src/daq/src/WaveformAnalysisRSNNLS.cc
```diff
@@ -0,0 +1,421 @@
...
+
+std::vector<std::pair<int, int>> WaveformAnalysisRSNNLS::FindThresholdRegions(const std::vector<double>& voltWfm,
+                                                                              double threshold) {
+  std::vector<std::pair<int, int>> regions;
+  bool in_region = false;
+  int region_start = -1;
+  const int padding = 1;
+
+  for (size_t i = 0; i < voltWfm.size(); ++i) {
+    if (voltWfm[i] < threshold) {
```
JamesJieranShen, line 299: These two nested if statements are just doing falling edge and rising edge detection, can we reduce the layers by just `and`ing the conditions?

src/daq/src/WaveformAnalysisRSNNLS.cc
```diff
@@ -0,0 +1,421 @@
...
+    }
+    return S;
+  };
+
+  // Iterative thresholding
+  int iter = 0;
+  const int max_iter = std::min(50, K);
+  int local_iterations_ran = 0;
+
+  while (!P.empty() && iter++ < max_iter) {
```
JamesJieranShen, line 230: Can we please rewrite this as a for loop? Just break out of the loop if P is empty.

src/daq/src/WaveformAnalysisRSNNLS.cc
```diff
@@ -0,0 +1,421 @@
...
+    for (size_t jj = 0; jj < cols.size(); ++jj) {
+      int c = cols[jj];
+      for (int i = 0; i < W.GetNrows(); ++i) S(i, jj) = W(i, c);
+    }
+    return S;
+  };
+
+  // Iterative thresholding
+  int iter = 0;
+  const int max_iter = std::min(50, K);
```
JamesJieranShen, line 227: no magic numbers please;) Also, this should presumably be configurable.

src/daq/src/WaveformAnalysisRSNNLS.cc
```diff
@@ -0,0 +1,421 @@
...
+  int local_iterations_ran = 0;
+
+  while (!P.empty() && iter++ < max_iter) {
+    local_iterations_ran = iter;
+
+    // Find component with minimum weight
```
JamesJieranShen, line 233: Can we use `std::min_element`?

src/daq/src/WaveformAnalysisRSNNLS.cc
```diff
@@ -0,0 +1,421 @@
...
+    double residual = voltVec(i) - fitted(i);
+    chi2_sum += residual * residual;
+  }
+
+  int active_components = 0;
+  for (int j = 0; j < K; ++j) {
+    if (h_full(j) > 0.0) active_components++;
+  }
+  int dof = std::max(1, D - active_components);
+  chi2ndf = chi2_sum / dof;
```
JamesJieranShen, line 285: Let's not use class variables for intermediate values.. instead let's pass in mutable function parameters.

src/daq/src/WaveformAnalysisRSNNLS.cc
```diff
@@ -0,0 +1,421 @@
...
+    merged_regions.push_back(regions[0]);
+
+    for (size_t i = 1; i < regions.size(); ++i) {
+      if (regions[i].first <= merged_regions.back().second + padding) {
+        merged_regions.back().second = regions[i].second;
+      } else {
+        merged_regions.push_back(regions[i]);
+      }
+    }
+    regions = merged_regions;
```
JamesJieranShen, line 329: use `std::swap` saves a copy

src/util/src/NNLS.cc
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  // Use SVD for underdetermined systems
+  if (nrows < ncols) {
+    TDecompSVD svd(AP);
+    Bool_t ok = kFALSE;
+    TVectorD temp_z = svd.Solve(b, ok);
+    if (!ok) throw std::runtime_error("NNLS: least-squares solve failed (SVD for underdetermined).");
+
+    // Manual copy for numerical stability
```
JamesJieranShen, line 44: I don't understand why a copy would ensure numerical stability, but from the code it seems that one can `std::swap(z, temp_z)` or just return `temp_z`?

src/util/src/NNLS.cc
```diff
@@ -0,0 +1,257 @@
...
+    double wmax = tol;  // Only consider gradients above tolerance
+    for (int j = 0; j < n; ++j) {
+      if (!inP[j] && w[j] > wmax) {
+        wmax = w[j];
+        t = j;
+      }
+    }
+
+    // Check termination conditions
+    if (t < 0 || outer_iter++ >= max_outer) break;
```
JamesJieranShen, line 137: convert this condition to a for loop please.

src/util/src/NNLS.cc
```diff
@@ -0,0 +1,257 @@
...
+    // Add variable t to passive set
+    inP[t] = 1;
+    Pidx.clear();
+    Pidx.reserve(n);
+    for (int j = 0; j < n; ++j) {
+      if (inP[j]) Pidx.push_back(j);
+    }
+
+    // Inner loop: solve unconstrained problem on passive set with feasibility enforcement
+    while (true) {
```
JamesJieranShen, line 148: I might be missing something, but it's not immediately obvious to me how it is guaranteed that this loop doesn't run indefinitely.

PR #338, opened 2025-09-23
Ensure MCPhotons are sorted by FrontEndTime in ForcedTrigger

JamesJieranShen:
MC Photons are added by their hit time (without PMT TTS). Without additional sorting by FrontEndTime, `DS::PMT` classes may not necessarily register the earliest photo electron in the event window. This leads to over-sampling of the late pulsing peak, as late pulses may have an early hit time.


tannerbk:
Thanks LGTM.

PR #339, opened 2025-09-25
Clean up digitNhitsCleaned calculation

hbjamin:
Remove unused digitpmt variable

PR #341, opened 2025-09-25
Concretize nhit logic

JamesJieranShen:
Previously it wasn't clear what is an `nhit` means for digitPMTs. This PR creates two new functions: `DigitNhits` and `DigitNhitsCleaned` in `DS::EV` to clarify that a channel contributes toward nhit if it has `DigitNCrossings() > 0`. This shouldn't impact any files created with zero suppression turned on, but would be useful if zero suppression is turned off.

Thanks to @hbjamin 's discussion on this matter:) This PR potentially supersedes #339 .


tannerbk:
LGTM.

PR #345, opened 2025-10-02
Updated Cerenkov class to handle materials with a non-monotonic refractive index.

ZachL00:
Screenshot 2025-10-17 at 9 32 19 AM
Mean photon yields from 1,000 events at each energy, from 1 to 10 MeV, in both LAr and water (error bars represent +- 1-StdDev from the mean). Old Cerenkov yields are shown as green stars, and the new yields from RAT::G4CerenkovProcess are shown as blue error bars. The Cerenkov yield in water is unchanged by this update (as expected, since it has a monotonic index), and the LAr yield follows a closer trend to the prediction especially as we approach lower energies (<4MeV). The yield in LAr is also higher now that we can properly integrate the refractive index below the scintillation peak wavelength (~128 nm).


src/physics/src/G4CerenkovProcess.cc
```diff
@@ -0,0 +1,615 @@
...
+  G4MaterialPropertiesTable* aMaterialPropertiesTable = aMaterial->GetMaterialPropertiesTable();
+
+  G4MaterialPropertyVector* Rindex = nullptr;
+
+  if (aMaterialPropertiesTable) Rindex = aMaterialPropertiesTable->GetProperty(kRINDEX);
+  // loop over refractive index vectors, 30-Sep-2025
+  // previous version used nMax = Rindex->GetMaxValue(), --> global maximum
+  G4double nMax;
+  if (Rindex) {
+    std::vector<G4double>* rindex_vals = rindex_vals = &(*refractiveIndexValuesVectors)[materialIndex];
```
tannerbk: Why are these set this way? Causes compiler warnings.

ZachL00: Hi Tanner, to follow up on this again I've removed the second assignment of each vector, as later in the class I call them correctly. This should address the compiler warnings you mentioned. Please let me know if there's anything else 👍

gmwendel:
PostStepGetPhysicalInteractionLength and PostStepDoIt are invoked on every single step of every charged particle.

Running std::minmax_element over the material's refractive index array multiple times per step to find nMin and nMax could slow the sim down.

Precomputing nMin and nMax for each material during initialization would solve this.
In G4CerenkovProcess.hh, add two new vectors

std::vector<G4double> fMinRindex;
std::vector<G4double> fMaxRindex;

In BuildPhysicsTable, after populating a material's refractive index array, find and store the nMin and nMax in these arrays.

In the stepping and DOIT methods, replace std::minmax_element with lookup:
G4double nMax = fMaxRindex[materialIndex];

ZachL00:
@gmwendel I've moved the computation of nMax and nMin to the BuildPhysicsTable function in the source file and verified that it works as expected. I ran 5,000 events in the Validation geometry, with both LAr and validwater, and see little difference before and after the update.

Screenshot 2026-04-13 at 7 41 40 PM
Produced Cherenkov yields in LAr.

Screenshot 2026-04-13 at 7 42 17 PM
Produced Cherenkov yields in validwater.

tannerbk:
Look good. @gmwendel please have a quick look when you get a chance.

PR #360, opened 2025-12-16
Add event time classifier

llebanowski:
This PR adds an event classifier that uses hit time residuals to provide:

1.  A ratio of hits in a time window (usually narrow and prompt)
2.  Four central moments of the values within a time window

Adds documentation and a new CLASSIFIER.ratdb table.
Needs PR #357 to write to ntuple.


azummo:
I agree, this PR looks ready to merge

PR #361, opened 2026-01-12
GLG4Scint: minor mods

llebanowski:
Move a conditional so that the subsequent conditional has a chance to be evaluated.


tannerbk:
Looks fine.

PR #364, opened 2026-01-13
Add complete LAB-PPO scintillator model

llebanowski:
This model is based on published parameters from the SNO+ collaboration, as well as the JUNO collaboration.  

depends on #362


tannerbk:
LGTM.

PR #376, opened 2026-02-24
Remove absorption from PMT glass models

llebanowski:
This PR adds new versions of the four PMT glasses currently in ratpac2.  The new versions have negligible optical absorption.  This is because quantum efficiency measurements inherently include glass absorption (confirmed with Hamamatsu), and therefore PMT models should not include glass absorption. 
Thus, this PR also updates all PMT models to use the new glasses.  

This PR also improves the refractive index of sapphire.  

The change to glasses could introduce a significant impact when using more green sensitive PMTs, but for the PMT in the Valid.geo, there is no observable impact with Cherenkov light:

No glass absorption:
image

With glass absorption:
Screenshot 2026-02-24 15 37 33

### DIFF
--- a/ratdb/OPTICS.ratdb
+++ b/ratdb/OPTICS.ratdb
@@ -234,6 +234,24 @@
 
 {
 "name": "OPTICS",
+"index": "glass",
+"run_range" : [0, 0],
+"surface": 1,
+"finish": "ground",
+"model": "glisur",
+"type": "dielectric_dielectric",
+"polish": 0.9,
+"RINDEX_option": "wavelength",
+"RINDEX_value1": [60.0, 200.0, 300.0, 600.0, 800.0, ],
+"RINDEX_value2": [1.458, 1.458, 1.458, 1.458, 1.458, ],
+"ABSLENGTH_option": "wavelength",
+"ABSLENGTH_value1": [60.0, 200.0, 300.0, 330.0, 500.0, 600.0, 770.0, 800.0, ],
+"ABSLENGTH_value2": [0.1e-3, 0.1e-3, 0.1e-3, 1.0e3, 2.0e3, 1.0e3, 0.5e3, 0.1e-3, ],
+"PROPERTY_LIST": ["RINDEX", "ABSLENGTH", ]
+}
+
+{
+"name": "OPTICS",
 "index": "ZnO2",
 "run_range" : [0, 0],
 "surface": 1,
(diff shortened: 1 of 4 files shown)


ratdb/OPTICS.ratdb
```diff
@@ -322,17 +317,21 @@
 "model": "glisur",
 "type": "dielectric_dielectric",
 "polish": 0.9,
+// Refractive index of Sapphire (Al2O3) from https://refractiveindex.info/?shelf=3d&book=crystals&page=sapphire {no data below 200 nm}
```
tannerbk: Can you make a separate PR for this change as it appears unrelated to the removal of the absorption lengths for the PMT glass.

ratdb/OPTICS.ratdb
```diff
@@ -366,10 +365,10 @@
 "RINDEX_option": "wavelength",
 "RINDEX_value1": [60.0, 200.0, 300.0, 600.0, 800.0, ],
 "RINDEX_value2": [1.458, 1.458, 1.458, 1.458, 1.458, ],
-"ABSLENGTH_option": "wavelength",
-"ABSLENGTH_value1": [60.0, 200.0, 300.0, 330.0, 500.0, 600.0, 770.0, 800.0, ],
-"ABSLENGTH_value2": [0.1e-3, 0.1e-3, 0.1e-3, 1.0e3, 2.0e3, 1.0e3, 0.5e3, 0.1e-3, ],
-"PROPERTY_LIST": ["RINDEX", "ABSLENGTH", ]
+//"ABSLENGTH_option": "wavelength",
```
tannerbk: Can these materials be used for other applications besides the PMTs? Won't this impact other uses of glass?

ratdb/OPTICS.ratdb
llebanowski: Done

ratdb/OPTICS.ratdb
llebanowski: In principle yes, though these definitions are placed under "PMT Materials".   In particular, "hamamatsu_borosilicate_glass" should not be used elsewhere, but "glass" could be.  
I think it makes sense to exclude "glass" as a PMT material and let it keep its absorption length.  The latest commit does this.  
As "glass" is currently used by "lappd", I have also switched it to use "borosilicate_glass", which should be more accurate.

PR #382, opened 2026-04-02
Support FOM lookup by both base and full fitter name in OutNtuple

JamesJieranShen:
Add GetFitterNameFromFullName helper to FitResult and use it in OutNtupleProc to merge FOMs declared under the base fitter name (e.g. fitter) with those declared under the full tagged name (e.g. `fitter__tag`), so tagged fitter variants inherit shared FOMs.

### DIFF
--- a/src/io/src/OutNtupleProc.cc
+++ b/src/io/src/OutNtupleProc.cc
@@ -20,6 +20,7 @@
 #include <RAT/OutNtupleProc.hh>
 #include <iostream>
 #include <numeric>
+#include <set>
 #include <sstream>
 #include <stlplus/string_utilities.hpp>
 #include <string>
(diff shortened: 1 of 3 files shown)


tannerbk:
This could use documentation.

azummo:
This looks good once the documentation is added

src/io/src/OutNtupleProc.cc
```diff
@@ -80,11 +81,25 @@ OutNtupleProc::OutNtupleProc() : Processor("outntuple") {
   event_fitters = table->GetSArray("event_fitters");
   for (const std::string &full_name : event_fitters) {
     if (event_fitter_FOMs.find(full_name) != event_fitter_FOMs.end()) continue;
+    std::set<std::string> fom_set;
+    // Load FOMs declared under the base fitter name (applies to all tagged variants)
+    std::string fitter_name = DS::FitResult::GetFitterNameFromFullName(full_name);
+    if (fitter_name != full_name) {
+      try {
+        for (const std::string &fom : table->GetSArray("event_fitter_FOM_" + fitter_name)) {
```
llebanowski, line 93: Minor suggestion: this try could be placed in the catch of the following try.  Thus, FOMs specified for a specific fitter (full_name) would take priority and only if not found would cause the default FOMs defined for the base fitter (fitter_name) to be saved.  
As it is, the default FOMs defined for the base fitter (fitter_name) will always be saved, which may not be desired.

JamesJieranShen:
Documentation is added, and previous comment addressed. @azummo @llebanowski please take a look. Since many of Logan's PRs depend on this, I will go ahead and merge this today unless there are additional concerns.

PR #384, opened 2026-04-08
bugfix(io): Prevent OutNtupleProc segfault on multiple beamOn commands

gmwendel:
Simulation crashes with a segmentation violation if multiple `/run/beamOn` commands are issued in a single macro while the `outntuple` processor is active.

**Root Cause:**
During run teardown, `OutNtupleProc::EndOfRun()` closes and deletes `outputFile`, but fails to nullify the pointer. Upon a subsequent `/run/beamOn`, the next `DSEvent()` call bypasses file recreation (`if (!this->outputFile)` evaluates false) and invokes `outputTree->Fill()` on a dangling pointer, resulting in a segfault. 

**Immediate Bugfix (this PR)**
Explicitly forbid multiple runs to prevent silent data loss from file overwriting. 
* Implement a `file_closed` boolean flagged during `EndOfRun()`. 
* If `DSEvent()` is called while `file_closed == true`, invoke `Log::Die()` with an error instructing the user cannot use ntuple with multiple beamon commands.

**Future work: Implement Full Multi-Run Support**
Refactor the ntuple IO architecture to safely accumulate data across runs. Very roughly this would require:
1. Explicitly nullify `outputFile`, `outputTree`, `metaTree`, and `waveformTree` in `EndOfRun()`.
2. Modify `OpenFile()` to use `"UPDATE"` mode instead of `"RECREATE"` for subsequent runs.
3. Add a `runId` branch to the `output` tree. This is strictly required so downstream analysis can perform relational joins to the correct run context in the `meta` tree.
4. existence checks in `OpenFile()` (e.g., `f->Get("output")`) to fetch and append to existing trees rather than instantiating new `TTree` cycles in the ROOT directory.

### DIFF
--- a/src/io/include/RAT/OutNtupleProc.hh
+++ b/src/io/include/RAT/OutNtupleProc.hh
@@ -52,7 +52,7 @@ class OutNtupleProc : public Processor {
   virtual void AssignAdditionalMetaAddresses(){};
   virtual void FillEvent(DS::Root *, DS::EV *){};
   virtual void FillNoTriggerEvent(DS::Root *){};
-  virtual void FillMeta() {};
+  virtual void FillMeta(){};
 
   // Exposed members for external tools
   DS::Run *runBranch;
--- a/src/io/include/RAT/OutNtupleProc.hh
+++ b/src/io/include/RAT/OutNtupleProc.hh
@@ -48,11 +48,11 @@ class OutNtupleProc : public Processor {
   virtual void EndOfRun(DS::Run *run) override;
 
   // Extensible functions
-  virtual void AssignAdditionalAddresses(){};
-  virtual void AssignAdditionalMetaAddresses(){};
-  virtual void FillEvent(DS::Root *, DS::EV *){};
-  virtual void FillNoTriggerEvent(DS::Root *){};
-  virtual void FillMeta(){};
+  virtual void AssignAdditionalAddresses() {};
+  virtual void AssignAdditionalMetaAddresses() {};
+  virtual void FillEvent(DS::Root *, DS::EV *) {};
+  virtual void FillNoTriggerEvent(DS::Root *) {};
+  virtual void FillMeta() {};
 
   // Exposed members for external tools
   DS::Run *runBranch;
@@ -77,6 +77,7 @@ class OutNtupleProc : public Processor {
   std::map<std::string, std::vector<std::string>> event_fitter_FOMs;
 
  protected:
+  bool has_run_completed;
   std::string defaultFilename;
   TFile *outputFile;
   TTree *outputTree;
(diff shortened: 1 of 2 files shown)


JamesJieranShen:
Hi @gmwendel , I don't think the full description of the PR has been applied as the pointers are not cleared in EndOfRun. I also don't necessarily think it is useful to have a separate bool flag (we can just use the original check on the pointer being null), and the fact that if nothing is done and multiple beamOn is called, ROOT will complain and die on the file-opening stage, as it will try to open a file that already exists.

gmwendel:
Good catch on the pointers, I intended pointer state management across multiple runs outside the scope of what I have time for here. This PR is a temporary fail-safe until someone has time to write an appropriate implementation. 

Using "RECREATE", I think an unhandled second beamOn will silently overwrite the first run if we are not careful. The explicit Log::Die prevents both the segfault and the silent data loss. I can PR description to reflect this narrower scope, but a separate issue should be opened for a true fix (since we claim in our docs multiple beamon commands can be run)

JamesJieranShen:
> Good catch on the pointers, I intended pointer state management across multiple runs outside the scope of what I have time for here. This PR is a temporary fail-safe until someone has time to write an appropriate implementation.
> 
> Using "RECREATE", I think an unhandled second beamOn will silently overwrite the first run if we are not careful. The explicit Log::Die prevents both the segfault and the silent data loss. I can PR description to reflect this narrower scope, but a separate issue should be opened for a true fix (since we claim in our docs multiple beamon commands can be run)



> Good catch on the pointers, I intended pointer state management across multiple runs outside the scope of what I have time for here. This PR is a temporary fail-safe until someone has time to write an appropriate implementation.
> 
> Using "RECREATE", I think an unhandled second beamOn will silently overwrite the first run if we are not careful. The explicit Log::Die prevents both the segfault and the silent data loss. I can PR description to reflect this narrower scope, but a separate issue should be opened for a true fix (since we claim in our docs multiple beamon commands can be run)

Sure, but I don't think this is a good place to check if RAT is simulating more than one run -- I also expect a lot more things to break than this processor if multiple runs are done, such as which run RATDB is looking at, stuff like that. So adding this check in the ntuple seems a bit strange. A coompromise may be to just issue a warning if RAT is trying to recreate a TFile for a existing path.

gmwendel:
If it is the case that ratpac will not support multiple beamon commands generally, I agree that the check should be done elsewhere.  This should be good to close, but an issue be raised to implement a feature that checks for multiple beamon commands and stops the simulation.

JamesJieranShen:
Please add the commits to set the pointers to null:)

tannerbk:
Closing for now, as rp2 does not support multiple beamOn commands.

PR #386, opened 2026-04-15
Fix RAVEN fitting no PEs for some threshold crossings

rpitelka:
RAVEN can fit zero PEs for a threshold crossing if the weight threshold is set above the voltage threshold. We want to always fit at least one PE for each threshold crossing.


JamesJieranShen (approved):

tannerbk:
Looks good.

PR #387, opened 2026-04-15
Modify FFTW3 configuration

JamesJieranShen:
as required for pick up FNAL UPS installs of FFTW on SL7.


tannerbk:
Have you tested this on other OSs?

JamesJieranShen:
Yes. Works with rp setup

PR #392, opened 2026-04-21
SplitEVDAQ: Add max_hit_duration

llebanowski:
The proposed `max_hit_duration` is very similar to `max_hit_time`.  The difference is that it does not start its time window at the beginning of a simulated event, but at the first hit.  This subtle difference will produce no change for most simulations, but can be important for delayed events with preceding events that do not produce hits.  One example is the beta decay of `Cs137` in `decay0`, where the beta may not enter the detector target while the delayed gamma may.  When using `max_hit_time`, the hits from the gamma are too late to be considered for a trigger.

### DIFF
--- a/src/daq/src/SplitEVDAQProc.cc
+++ b/src/daq/src/SplitEVDAQProc.cc
@@ -26,6 +26,7 @@ void SplitEVDAQProc::BeginOfRun(DS::Run *run) {
   fTriggerResolution = ldaq->GetD("trigger_resolution");
   fLookback = ldaq->GetD("lookback");
   fMaxHitTime = ldaq->GetD("max_hit_time");
+  fMaxHitDuration = ldaq->GetD("max_hit_duration");
   fTriggerOnNoise = ldaq->GetI("trigger_on_noise");
   fDigitizerType = ldaq->GetS("digitizer_name");
   fDigitize = ldaq->GetZ("digitize");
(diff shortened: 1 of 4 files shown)


src/daq/src/SplitEVDAQProc.cc
```diff
@@ -78,7 +79,8 @@ Processor::Result SplitEVDAQProc::DSEvent(DS::Root *ds) {
   start = floor(start / fTriggerResolution) * fTriggerResolution;
   double end = *std::max_element(trigPulses.begin(), trigPulses.end());
   end = (floor(end / fTriggerResolution) + 1) * fTriggerResolution;
-  std::sort(trigPulses.begin(), trigPulses.end());
```
tannerbk, line 81: Why is this sort removed?

src/daq/src/SplitEVDAQProc.cc
```diff
@@ -78,7 +79,8 @@ Processor::Result SplitEVDAQProc::DSEvent(DS::Root *ds) {
...
+
+  if (fMaxHitDuration > 0 && end - start > fMaxHitDuration) end = start + fMaxHitDuration;
```
tannerbk: I wonder if it is worth warning that there are hits outside the window that will not be recorded in a trigger.

Also does this not risk extremely long loops where `nbins` is a very large number? For example, imagine start = 0  and we have a long-live state that decays 1 second later, so end ~ 1 second. 

(also, I'd wrap (end - start) in parentheses for clarity).

src/daq/src/SplitEVDAQProc.cc
llebanowski: The only place trigPulses is used after this sort is in a for loop over all elements of trigPulses and in that loop the value each element is used directly in a calculation - the order does not matter.

src/daq/src/SplitEVDAQProc.cc
llebanowski: A warning sounds like a nice idea though it could be spammy if warning on every event.  Open to suggestions.  

The risk due to this change is equal to the existing risk because this change only shifts a zero time and does not change the window length.  I did test a case that had many more bins and it was proportionally slower.  Ultimately, the user should check the DAQ settings.  

Added parentheses.

tannerbk:
I think this looks fine. Just needs minor update to docs.

PR #397, opened 2026-05-14
Add the elliptical tube factory to RAT

vlozza:


JamesJieranShen (approved):
Thanks @vlozza ! Seems useful.

PR #398, opened 2026-05-18
Upgrade basic fitter options

llebanowski:
- Add "light_speed" to "FIT_COMMON" table, and remove it from a couple of individual tables.
- Enable selection of hits by pmt_type in quadFitter and fitCentroid
- Add fit label to quadFitter and fitCentroid
- Add Cartesian limits to quadFitter positions (optional alternative to radial limit)
- Add PMT hit time range option to quadFitter


ratdb/FITTER.ratdb
```diff
@@ -7,6 +7,7 @@
 "vertex_seed": "quadfitter",
 "direction_seed": "fitdirectioncenter",
 "energy_seed": "DOESNOTEXIST"
```
azummo: Missing a comma here

src/fit/src/FitQuadProc.cc
```diff
@@ -223,14 +267,14 @@ Processor::Result FitQuadProc::Event(DS::Root *ds, DS::EV *ev) {
...
         // This is meant to remove wild fits, not directly
         // restrict them to the physically allowed region.
-        // Indeed, if we cut this off at the PMT sphere surface,
+        // Indeed, if we cut this off at the PMT surface,
         // it would introduce a large bias when we took the
-        // median later.  So it is a little larger than that.
+        // median later.  So it should be a little larger than that.
         const double rlimit = fMaxRadius;
-        if (mag2(v) < rlimit * rlimit) {
+        if (mag2(v) < rlimit * rlimit || (abs(v[0]) < fMaxX && abs(v[1]) < fMaxY && abs(v[2]) < fMaxZ)) {
```
azummo, line 303: What happens if you only set one of the max x/y/z? I think this will always fail in that case. Should only check that position < fMaxX/Y/Z when fMaxX/Y/Z is set, or require all 3 are set before checking.

azummo:
Do these changes require any updates to documentation for the fitters?

src/fit/src/FitQuadProc.cc
```diff
@@ -127,22 +152,46 @@ static inline int matinvert(double (*const ans)[3], const double (*const m)[3])
 }
 
 Processor::Result FitQuadProc::Event(DS::Root *ds, DS::EV *ev) {
+  if (fMaxRadius > 0) {
+    if (fMaxX > 0 || fMaxY > 0 || fMaxZ > 0)
+      Log::Die("Quad tried to set both max_radius and (max_x or max_y or max_z).");
```
azummo: I get this error whenever setting max_x/y/z, even without setting max_radius. Is it still using the default max radius value?

src/fit/src/FitQuadProc.cc
```diff
@@ -127,22 +152,46 @@ static inline int matinvert(double (*const ans)[3], const double (*const m)[3])
...
+      }
+      if (iType == fPMTtype.size()) continue;  // No match found
+    }
+
     double time = inputHandler.GetTime(pmtid);
     if (time > 1e6) {
       continue;
     }
+    // Select PMTs by time - optional
+    if (fMaxHitTime > fMinHitTime && (time > fMaxHitTime || time < fMinHitTime)) {
```
azummo: This will only work if both Max and Min are set. If the user tries to set only one, it will run but not actually cut any hits.

llebanowski: fixed by introducing two booleans

src/fit/src/FitQuadProc.cc
llebanowski: fixed by adding max_x/y/z to FIT_QUAD.ratdb.

azummo: I'm still getting the same error after these changes.

tannerbk: @llebanowski can you address this?

azummo:
Looks good and is working for me now.

Issue #391, opened 2026-04-20
Build failure on macOS/Clang: errors raised on macOS but not Linux


mdcge:
Building ratpac-two on macOS (M4 chip, Sequoia 15.5) fails with these errors:

1.
```
error: cannot pass object of non-trivial type 'std::string' (aka 'basic_string<char>') through variadic function; call will abort at runtime [-Wnon-pod-varargs]
  return dformat("Using wavelength specrum:\t%d\t%s", fNumPhotons, fWavelengthIndex);
```
`src/gen/src/VertexGen_PhotonBomb.cc`, line 208

2.
```
error: use of overloaded operator '<<' is ambiguous (with operand types 'otext' and 'rep' (aka 'long long'))
  debug << "Deconvolution took " << std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count() << " ms."
```
`src/daq/src/WaveformAnalysisLucyDDM.cc`, line 62, 74 & 164

Both of these don't appear on Linux but do on macOS, this is probably because Clang treats these as errors, whereas GCC treats them as warnings.

I believe the fixes should be as simple as:

1. Use `fWavelengthIndex.c_str()` instead of `fWavelengthIndex`.
2. Use `static_cast<long>(std::chrono::duration_cast<std::chrono::milliseconds>(end - start.count())` for lines 62 and 74. Use `static_cast<long>(std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - start).count())` for line 164.

These errors could also be fixed by adding the `-Wnon-pod-varargs` flag to the CXX flags (although this could be dangerous as it would silence other potentially important warnings).

Also, "specrum" is a typo.

tannerbk:
@mdcge please feel free to submit a PR with these changes.

JamesJieranShen:
fixed by #429

PR #399, opened 2026-05-18
Fix MacOS/Clang error by explicitly casting to string

mdcge:
Fixes issue #391.

These lines cause errors when building on MacOS due to Clang treating these non-explicit strings as errors, whereas GCC accepts them with a warning.

Clang requires variables in specific cases (for example when using the `<<` operator) to be explicitly cast to string.

Also corrected the typo "specrum" -> "spectrum".

### DIFF
--- a/src/daq/src/WaveformAnalysisLucyDDM.cc
+++ b/src/daq/src/WaveformAnalysisLucyDDM.cc
@@ -59,7 +59,8 @@ void WaveformAnalysisLucyDDM::DoAnalysis(DS::DigitPMT* digitpmt, const std::vect
   size_t iterations_ran = 0;
   std::vector<double> demod_result = Deconvolve(voltWfm, iterations_ran);
   auto end = std::chrono::high_resolution_clock::now();
-  debug << "Deconvolution took " << static_cast<int64_t>(std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count()) << " ms."
+  debug << "Deconvolution took "
+        << static_cast<int64_t>(std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count()) << " ms."
         << newline;
   std::vector<double> reblurred_wfm = ReblurWaveform(demod_result);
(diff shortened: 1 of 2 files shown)


src/gen/src/VertexGen_PhotonBomb.cc
```diff
@@ -205,7 +205,7 @@ void VertexGen_PhotonBomb::SetState(G4String newValues) {
 
 G4String VertexGen_PhotonBomb::GetState() {
   if (fSpectrum)
-    return dformat("Using wavelength specrum:\t%d\t%s", fNumPhotons, fWavelengthIndex);
+    return dformat("Using wavelength spectrum:\t%d\t%s", fNumPhotons, fWavelengthIndex.str());
```
JamesJieranShen: The CI caught a typo here. Did you mean `c_str`?

src/daq/src/WaveformAnalysisLucyDDM.cc
```diff
@@ -59,7 +59,7 @@ void WaveformAnalysisLucyDDM::DoAnalysis(DS::DigitPMT* digitpmt, const std::vect
   size_t iterations_ran = 0;
   std::vector<double> demod_result = Deconvolve(voltWfm, iterations_ran);
   auto end = std::chrono::high_resolution_clock::now();
-  debug << "Deconvolution took " << std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count() << " ms."
+  debug << "Deconvolution took " << static_cast<long>(std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count()) << " ms."
```
JamesJieranShen: Prefer using fixed-size integers (`int64_t`), as `long` can have different lengths in different systems.

JamesJieranShen (changes requested):
Thanks @mdcge for preparing this! See below for some very minor requests, as well as a typo caught by the CI.

PR #400, opened 2026-05-18
Fix typo in stainless_steel and add two more

llebanowski:
Adds two more common stainless steels and fixes an apparent typo in the "stainless_steel" definition, which could have a broad impact on users.


tannerbk:
LGTM

PR #402, opened 2026-05-27
Fit valid flag for waveform analyzers

rpitelka:
Add a fit valid flag for waveform analyzers that is false if the analyzer is not run or does not fit a PE to a DigitPMT. Prevents crashes when event reconstruction algorithms are run in waveform analyzer mode.

### DIFF
--- a/doc/users_guide/outputs.rst
+++ b/doc/users_guide/outputs.rst
@@ -180,6 +180,7 @@ If ``include_digitizerfits`` is set then we additionally add the following infor
 ``fit_time_+{fitter_name}``             vector<double>       The time extracted from each PMT waveform fit.
 ``fit_charge_+{fitter_name}``           vector<double>       The charge extracted from each PMT waveform fit.
 ``fit_FOM_+{fitter_name}_+{fom_name}``  vector<double>       The figure of merit extracted from each PMT waveform fit.
+``fit_valid_+{fitter_name}``            vector<bool>         Per-PMT flag (parallel to ``digitPMTID``) indicating whether the named fitter produced a valid result for that PMT. ``false`` when the fitter did not run on that PMT or its analysis reported no extracted PEs.
 ======================================  ===================  ===================
 
 If ``include_nestedtubehits`` is set then we additionally add the following information to the ``output`` branch of the ntuple. These "nested tubes" are intended for use with liquid-O style fiber optics detectors. These are filled from the ``DS::MCNestedTube`` branch.
--- a/src/ds/include/RAT/DS/WaveformAnalysisResult.hh
+++ b/src/ds/include/RAT/DS/WaveformAnalysisResult.hh
@@ -36,6 +36,7 @@ class WaveformAnalysisResult : public TObject {
       std::vector<Double_t>& fom_array = figures_of_merit[kv.first];
       fom_array.insert(fom_array.begin() + insertion_index, kv.second);
     }
+    fit_valid = true;
     return insertion_index;
   }
   virtual void setTimeOffset(Double_t _offset) { time_offset = _offset; }
@@ -47,7 +48,13 @@ class WaveformAnalysisResult : public TObject {
 
   virtual const std::vector<Double_t>& getTimes() { return times; }
   virtual const std::vector<Double_t>& getCharges() { return charges; }
-  ClassDef(WaveformAnalysisResult, 2);
+
+  virtual void setFitValid(Bool_t v) { fit_valid = v; }
+  // The `|| !times.empty()` clause keeps older ROOT files readable: pre-v3 results
+  // serialize fit_valid as default-false, but a non-empty PE vector still implies validity.
+  virtual Bool_t getFitValid() const { return fit_valid || !times.empty(); }
+
+  ClassDef(WaveformAnalysisResult, 3);
 
  protected:
   // All arrays are parallel arrays. It is assumed that they are sorted in time
@@ -55,6 +62,7 @@ class WaveformAnalysisResult : public TObject {
   std::vector<Double_t> charges;
   std::map<std::string, std::vector<Double_t>> figures_of_merit;
   Double_t time_offset;
+  Bool_t fit_valid = false;
 };
 
 }  // namespace DS
--- a/src/fit/include/RAT/FitterInputHandler.hh
+++ b/src/fit/include/RAT/FitterInputHandler.hh
@@ -372,6 +372,22 @@ class FitterInputHandler {
     }
   }
 
+  /**
+   * @brief Check whether the configured waveform analyzer produced a valid result for a PMT.
+   *
+   * In kWaveformAnalysis mode, returns the analyzer's fit_valid flag — false for empty results
+   * (e.g. RAVEN finding no threshold crossings) so callers can skip them safely. In kPMT and
+   * kDigitPMT modes the hit list itself is the validity filter, so this always returns true.
+   *
+   * @param id PMT ID.
+   */
+  bool IsHitValid(Int_t id) const {
+    if (mode != Mode::kWaveformAnalysis) return true;
+    if (!ev) return false;
+    DS::DigitPMT* digitpmt = ev->GetOrCreateDigitPMT(id);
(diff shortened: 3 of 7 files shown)


src/fit/include/RAT/FitterInputHandler.hh
```diff
@@ -372,6 +372,21 @@ class FitterInputHandler {
...
 
+  /**
+   * @brief Check whether the configured waveform analyzer produced a valid result for a PMT.
+   *
+   * Only meaningful in kWaveformAnalysis mode; other modes filter via the hit list and
+   * always return true.
+   *
+   * @param id PMT ID.
+   */
+  bool IsHitValid(Int_t id) const {
```
JamesJieranShen, line 383: Probably better to name this `IsFitValid`, since it is not per hit/npe.

src/fit/mimir/include/mimir/Cost.hh
```diff
@@ -21,6 +21,7 @@ class Cost {
   virtual void AddAllHits(const RAT::FitterInputHandler& input_handler) {
     std::vector<int> pmtids = input_handler.GetAllHitPMTIDs();
     for (int pmtid : pmtids) {
+      if (!input_handler.IsHitValid(pmtid)) continue;
```
JamesJieranShen, line 24: I don't  know if it is the best idea to patch this in here. MIMIR should allow invalid hits to be used during reconstruction, if the likelihood accepts it. Probably best to record the validities of the hits separately and let individual likelihoods decide what they would like to do.

src/io/include/RAT/OutNtupleProc.hh
```diff
@@ -211,6 +211,8 @@ class OutNtupleProc : public Processor {
   std::map<std::string, std::vector<double>> wfmFitTime;
   std::map<std::string, std::vector<double>> wfmFitCharge;
   std::map<std::string, std::map<std::string, std::vector<double>>> wfmFitFOM;
+  // Per-PMT fit validity, parallel to digitPMTID.
```
JamesJieranShen, line 214: Remove the comment here, since it is redundant given the comment in l209, and that the other maps are not equally commented.

src/io/src/OutNtupleProc.cc
```diff
@@ -632,9 +634,16 @@ Processor::Result OutNtupleProc::DSEvent(DS::Root *ds) {
         digitPeak.push_back(digitpmt->GetPeakVoltage());
         digitLocalTriggerTime.push_back(digitpmt->GetLocalTriggerTime());
         if (options.digitizerfits) {
-          const std::vector<std::string> fitters = digitpmt->GetFitterNames();
-          for (std::string fitter_name : fitters) {
+          const std::vector<std::string> ran_fitter_names = digitpmt->GetFitterNames();
+          std::set<std::string> ran_fitters(ran_fitter_names.begin(), ran_fitter_names.end());
+          for (const std::string &fitter_name : waveform_fitters) {
+            if (!ran_fitters.count(fitter_name)) {
+              // Fitter never ran on this PMT — record invalid, leave per-PE arrays untouched.
```
JamesJieranShen, line 641: Per discussion during the meeting, I think in `OutNtuple` this specific condition is worthy of a direct `die`, rather than a silent error, since this indicates a logic error on the waveform analyzer side.

src/io/src/OutNtupleProc.cc
```diff
@@ -632,9 +634,16 @@ Processor::Result OutNtupleProc::DSEvent(DS::Root *ds) {
         digitPeak.push_back(digitpmt->GetPeakVoltage());
         digitLocalTriggerTime.push_back(digitpmt->GetLocalTriggerTime());
         if (options.digitizerfits) {
-          const std::vector<std::string> fitters = digitpmt->GetFitterNames();
-          for (std::string fitter_name : fitters) {
+          const std::vector<std::string> ran_fitter_names = digitpmt->GetFitterNames();
+          std::set<std::string> ran_fitters(ran_fitter_names.begin(), ran_fitter_names.end());
```
JamesJieranShen, line 638: Having a set is unnecessary and likely slower, since the vector of strings is typically very small. Let's just do `std::find` on the vectors themselves.

src/ds/include/RAT/DS/WaveformAnalysisResult.hh
```diff
@@ -47,14 +48,21 @@ class WaveformAnalysisResult : public TObject {
 
   virtual const std::vector<Double_t>& getTimes() { return times; }
   virtual const std::vector<Double_t>& getCharges() { return charges; }
-  ClassDef(WaveformAnalysisResult, 2);
+
+  virtual void setFitValid(Bool_t v) { fit_valid = v; }
+  // The `|| !times.empty()` clause keeps older ROOT files readable: pre-v3 results
+  // serialize fit_valid as default-false, but a non-empty PE vector still implies validity.
+  virtual Bool_t getFitValid() const { return fit_valid || !times.empty(); }
```
JamesJieranShen, line 55: No need to really maintain backward compatibility in the RATDS structs. Furthermore, the OR here prevents invalid modes where fit is actually invalid despite times being present. I think a naive getter is fine here.

src/ds/include/RAT/DS/WaveformAnalysisResult.hh
```diff
@@ -47,14 +48,21 @@ class WaveformAnalysisResult : public TObject {
 
   virtual const std::vector<Double_t>& getTimes() { return times; }
   virtual const std::vector<Double_t>& getCharges() { return charges; }
-  ClassDef(WaveformAnalysisResult, 2);
+
+  virtual void setFitValid(Bool_t v) { fit_valid = v; }
```
JamesJieranShen, line 52: avoid one character variable names.

doc/users_guide/outputs.rst
```diff
@@ -180,6 +180,7 @@ If ``include_digitizerfits`` is set then we additionally add the following infor
 ``fit_time_+{fitter_name}``             vector<double>       The time extracted from each PMT waveform fit.
 ``fit_charge_+{fitter_name}``           vector<double>       The charge extracted from each PMT waveform fit.
 ``fit_FOM_+{fitter_name}_+{fom_name}``  vector<double>       The figure of merit extracted from each PMT waveform fit.
+``fit_valid_+{fitter_name}``            vector<bool>         Per-PMT flag (parallel to ``digitPMTID``) indicating whether the named fitter produced a valid result for that PMT. ``false`` when the fitter did not run on that PMT or its analysis reported no extracted PEs.
```
JamesJieranShen, line 183: Do we actually want this? Everything in `fit_***` is one-per-PE, not one-per-channel, including `fit_pmtid`. If we really want to maintain this on the per-channel level, I would suggest that this field would need to be named something like `digitFitValid_RAVEN`, rather than `fitXXX`.

src/io/src/OutNtupleProc.cc
```diff
@@ -632,9 +634,16 @@ Processor::Result OutNtupleProc::DSEvent(DS::Root *ds) {
...
+          for (const std::string &fitter_name : waveform_fitters) {
+            if (!ran_fitters.count(fitter_name)) {
+              // Fitter never ran on this PMT — record invalid, leave per-PE arrays untouched.
+              wfmFitValid[fitter_name].push_back(false);
+              continue;
+            }
             DS::WaveformAnalysisResult *fit_result = digitpmt->GetOrCreateWaveformAnalysisResult(fitter_name);
+            wfmFitValid[fitter_name].push_back(fit_result->getFitValid());
```
JamesJieranShen, line 646: See my comment on the doc about this.

JamesJieranShen (changes requested):
@rpitelka Thanks again for working on this! See some comments.

src/ds/include/RAT/DS/WaveformAnalysisResult.hh
rpitelka: I thought I deleted this comment lol

rpitelka:
Putting this on hold as it's not necessary for the upcoming coordination effort.

rpitelka:
Made redundant by PR #406 and #428. Current policy is that if a waveform analyzer is run, it is expected to produce at least 1 PE result per threshold-crossing region.

PR #403, opened 2026-05-29
Move creation of WaveformAnalysisResult to base class

rpitelka:
In order to enforce the principle that a `WaveformAnalysisResult` should always be written when a waveform analyzer is run on a `DigitPMT`, the creation of the result is moved to `WaveformAnalyzerBase`. This fixes a bug cause by the current implementation of min and max charge thresholds in the base class; when the total charge is outside of the specified charge window, `DoAnalysis` is not run, and no `WaveformAnalysisResult` is written.

### DIFF
--- a/src/daq/include/RAT/WaveformAnalyzerBase.hh
+++ b/src/daq/include/RAT/WaveformAnalyzerBase.hh
@@ -108,6 +108,14 @@ class WaveformAnalyzerBase : public Processor {
    */
   virtual void DoAnalysis(DS::DigitPMT *digitpmt, const std::vector<UShort_t> &digitwfm) = 0;
 
+  /**
+   * Short name under which this analyzer's WaveformAnalysisResult is stored (e.g. "Lognormal").
+   * Used to guarantee a result is always created, even when analysis is skipped (e.g. due to
+   * the charge thresholds). Must match the name passed to GetOrCreateWaveformAnalysisResult()
+   * inside DoAnalysis().
+   */
+  virtual std::string GetAnalyzerName() const = 0;
+
  protected:
   // Digitizer Settings
   double fTimeStep;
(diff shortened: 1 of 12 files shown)


src/daq/include/RAT/WaveformAnalyzerBase.hh
```diff
@@ -108,6 +108,14 @@ class WaveformAnalyzerBase : public Processor {
    */
   virtual void DoAnalysis(DS::DigitPMT *digitpmt, const std::vector<UShort_t> &digitwfm) = 0;
 
+  /**
+   * Short name under which this analyzer's WaveformAnalysisResult is stored (e.g. "Lognormal").
+   * Used to guarantee a result is always created, even when analysis is skipped (e.g. due to
+   * the charge thresholds). Must match the name passed to GetOrCreateWaveformAnalysisResult()
+   * inside DoAnalysis().
+   */
+  virtual std::string GetAnalyzerName() const = 0;
```
JamesJieranShen: Do we just want to use `GetName` from the processor class? That may be easier than adding another layer to all these labels.

JamesJieranShen (commented):
This seems fine, besides my minor comments.

One thing to consider, though, is that this PR further entrenches the fact that multiple instances of the same processor cannot be run in the same RAT macro. This has already been the case for most analyzers, but if we are settled on that decision (at least for now), perhaps we should have error checks when someone messes up and try to run multiple instances at the time. As far as I can tell, the result of one optimizer gets dropped silently if the code is left as-is.

src/daq/include/RAT/WaveformAnalyzerBase.hh
rpitelka: That returns the full name "WaveformAnalysisFitterName", but I can just strip away the beginning and then we won't have to override, which is def nicer.

PR #404, opened 2026-05-30
Create RAT::DS::RootFactory, allow RAT::DS data structures to be extended by private experiments

JamesJieranShen:
Instead of directly calling `RAT::DS::Root`'s constructor, invoke a call to a DS ROOT factory to allow a derived class to be provided to the processors, should one be implemented.

Also removes the `UserData` vector in the `RAT::DS::Root` class, as these seem less useful given the current data structure. I don't think anyone has ever used these.


JamesJieranShen:
Updated: now deprecated RATEvent, a library of minimal dependencies that rat's ROOT depends  on. Now, just link to the main libRATPAC library to allow for Geant4 routines to also be called. Useful to if geometry details are needed during analysis (e.g. to compute time residuals).

tannerbk:
Almost certainly this could use updated documentation. Overall changes look good, but I haven't had a chance to test yet.

JamesJieranShen:
Additional documentation has been added.

src/ds/include/RAT/DS/RootFactory.hh
```diff
@@ -0,0 +1,36 @@
+#include <TClass.h>
```
tannerbk, line 1: This is missing the standard comment structure in the RAT::DS files that I think is useful enough to maintain.

tannerbk:
Looks good to me, pending minor comment.

PR #406, opened 2026-06-01
Add warning to log when total charge threshold is exceeded

rpitelka:
Add warning to log when a waveform analyzer doesn't reconstruct due to the total charge on a PMT exceeding the set threshold.

### DIFF
--- a/src/daq/src/WaveformAnalyzerBase.cc
+++ b/src/daq/src/WaveformAnalyzerBase.cc
@@ -31,7 +31,11 @@ void WaveformAnalyzerBase::RunAnalysis(DS::DigitPMT* digitpmt, int pmtID, DS::Di
   digitpmt->GetOrCreateWaveformAnalysisResult(GetAnalyzerName());
 
   double totalCharge = digitpmt->GetDigitizedTotalCharge();
-  if (totalCharge < fMinTotalCharge || totalCharge > fMaxTotalCharge) return;
+  if (totalCharge < fMinTotalCharge || totalCharge > fMaxTotalCharge) {
+    warn << "Total charge " << totalCharge << " is outside of the allowed range [" << fMinTotalCharge << ", "
+         << fMaxTotalCharge << "]. Skipping waveform analysis for PMT " << pmtID << "." << newline;
+    return;
+  }
 
   fVoltageRes = dsdigit->GetVoltageResolution();
   fTimeStep = dsdigit->GetTimeStepNS();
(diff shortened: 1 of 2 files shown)


src/daq/src/WaveformAnalyzerBase.cc
```diff
@@ -19,7 +19,11 @@ void WaveformAnalyzerBase::RunAnalysis(DS::DigitPMT* digitpmt, int pmtID, Digiti
 
 void WaveformAnalyzerBase::RunAnalysis(DS::DigitPMT* digitpmt, int pmtID, DS::Digit* dsdigit) {
   double totalCharge = digitpmt->GetDigitizedTotalCharge();
-  if (totalCharge < fMinTotalCharge || totalCharge > fMaxTotalCharge) return;
+  if (totalCharge < fMinTotalCharge || totalCharge > fMaxTotalCharge) {
+    warn << "Total charge " << totalCharge << " is outside of the allowed range [" << fMinTotalCharge << ", "
```
tannerbk, line 23: I think this is way too spammy to warn every time this happens, which could in principle be every single hit. You could just warn once?

JamesJieranShen: also, seems more like an `info`/`debug` than a warn.

JamesJieranShen:
Dropping the printout level to info seems reasonable to me (it can be even further lowered to DEBUG imo, if we handle the max-charge case properly). I find the only warn once implementation a bit strange though -- if someone is debugging this, they would have no clue based on the print out which PMTs crossed the charge threshold (except for the first one).

rpitelka:
> Dropping the printout level to info seems reasonable to me (it can be even further lowered to DEBUG imo, if we handle the max-charge case properly). I find the only warn once implementation a bit strange though -- if someone is debugging this, they would have no clue based on the print out which PMTs crossed the charge threshold (except for the first one).

We expect muons to cross the threshold in data, which is the main use case for setting the threshold (to prevent waveform analyzers from slowing down processing), so I think keeping it at INFO make more sense. The point of this message as I envisioned it is 1) to be a way for users running simulation macros to quickly test that the threshold is not set too low, and 2) in data to let the user know that a very high charge event (like a muon) has occurred. The information about which PMTs crossed threshold can be obtained from the ntuples via `digitCharge` and the presence of an empty waveform analysis result.

JamesJieranShen (approved):

PR #410, opened 2026-06-14
Throw FatalError from Log::Die instead of std::exit

JamesJieranShen:
Log::Die() and a failed Log::Assert() now throw a RAT::FatalError carrying the return code rather than calling std::exit(). This allows the ROOT interpreter to stay alive when `Log::Die` is triggered.


tannerbk:
Looks good.

PR #412, opened 2026-06-14
Transittime calculation

JamesJieranShen:
Introduce the transit time calculator and friends.

The transit time calculator requires two components: a LightPathCalculator, and a GroupVelocityCalculator. LightPathCalcualtor is responsible for creating tracked "segments" of materials that a photon propagates through. The GroupVelocityCalculator then takes these segments, computes a group velocity for each segment. The main transittime calculator then aggregates the results into a full transit time calculation.

Both `LightPathCalculator` and `GroupVelocityCalculator` is dispatched via RAT's existing `GlobalFactory` mechanism, used for all the generators. One implementation is provided for each.
- The `G4LightPathCalculator` assumes a straight line from the start position to the end position, and utializes G4's navigation features to identify the length of the line in each amterial.
- The `RIndexGroupVelocityCalculator` computes the group velocity using the refractive index defined in the optical materials.

This PR also introduces a new field to the NTuples, `mcTransitTimeToPMTs`. For each event, this field is populated with a vector that computes the transit time from the event vertex (`mcx`, `mcy`, `mcz`) To each PMT (indexed by PMTID). 

Requires #404


tannerbk:
This needs documentation, and an example for using the calculator would be very helpful.

JamesJieranShen:
Added offline analysis macro + some improvements to the DSReader to make it more usable (and closer to SNO+'s implementation, which most people might be more familiar with). 

Below is the output of the offline analysis macro on 10k photon bombs uniformly distributed in the `validation` geometry.
image

tannerbk:
This worked fine for me. I think this still needs documentation, in particular how to include these methods into a RAT fitter.

src/io/src/DSReader.cc
```diff
@@ -1,46 +1,143 @@
...
 
 #ifdef DEBUG
   debug << "DSReader::Add - "
         << "filename='" << filename << "', total=" << total << newline;
 #endif
 }
 
-DS::Root *DSReader::NextEvent() {
-  if (next < total) {
+const DS::Root &DSReader::NextEntry() {
```
tannerbk: Changing the naming here can break many things for downstream experiments and analysis code. Indeed it breaks at least one ratpac-two example code.

JamesJieranShen: Potentially the old name can get confusing in my opinion, since we would have EV and Event meaning different things.. I think a DSReader improvement/upgrade is needed anyways, see #415 , so perhaps this is a good time to rename things to make the class less confusing?

tannerbk: I agree the new naming is better, or at least more in-line with the other RAT naming. The change breaks this example script: https://github.com/rat-pac/ratpac-two/blob/main/examples/tracking.py#L19 and it will need to be announced.

JamesJieranShen: This is an oversight on my part for not checking the python bindings. Just pushed a new commit. Can you try again?

I also agree that this needs to be announced.

src/io/include/RAT/DSReader.hh
```diff
@@ -5,41 +5,69 @@
...
+
+  void Add(const std::string &filename);
+  void SetBranchStatus(const std::string &bname, bool status = 1) { T.SetBranchStatus(bname.c_str(), status); };
 
   TTree *GetT() { return &T; };
   TTree *GetRunT() { return &runT; };
-  DS::Root *GetDS() { return ds; };
-  Long64_t GetTotal() { return total; };
+  virtual DS::Root *GetDS() { return fDS; };
+  Long64_t GetEntryCount() { return fTotalEntries; };
```
azummo, line 42: I think this has the same effect as the GetEvent->GetEntry change. See https://github.com/rat-pac/ratpac-two/blob/main/examples/tracking.py#L17

examples/tracking.py
```diff
@@ -16,7 +16,7 @@
 # Loop over the simulated events
 for ev in range(ds.GetTotal()):
```
azummo: This should be changed to ds.GetEntryCount()

src/ds/include/RAT/DS/RootFactory.hh
azummo, line 1: Should this have an include guard?

azummo:
I tested this, and it works for me. I agree with Tanner's comments on documentation. Only have minor comments above.

JamesJieranShen:
Thanks @azummo for the coments. They should be addressed now. 

On documentation -- there already is an example in OutNtupleProc on how the calculator can be used in a rat processor. Do we really need a specific one for a fitter? It is exactly the same call.

PR #413, opened 2026-06-30
Support for local NLopt and Tensorflow

scott-degraw:
The current CMakeLists don't give the targets any include directories for NLopt or Tensorflow so if they are not on global include paths it fails to build. 

The `Decay0` object was also not ROOT 6.40 compatible due to TF1 removing CINT [support](https://root.cern/doc/v640/release-notes.html).


CMakeLists.txt
```diff
@@ -141,7 +145,6 @@ include_directories(${XercesC_INCLUDE_DIRS})
 
 find_package(Threads REQUIRED)
 find_package(PkgConfig REQUIRED)
-find_package(FFTW3 CONFIG)
```
JamesJieranShen, line 144: Hi Scott, why is this line removed? I believe this line allowed fftw's native cmake to be pulled in, should it exist.

scott-degraw: Hi James, I forgot to comment on that part. The build was having problems with fftw using the lcg view at `/cvmfs/sft.cern.ch/lcg/views/LCG_110/x86_64-el9-gcc15-opt`. This seems to be related to this [issue](https://github.com/FFTW/fftw3/issues/130) fftw seems to have poor support for CMake so I just disabled it. The `pkg-config` works fine though so cmake is able to get the local fftw through the `PKG_CONFIG_PATH` variable.

JamesJieranShen (approved):
LGTM, and thanks -- we should now be able to move to ROOT v6.40 (the next long-term-support version of ROOT) in the next release.

JamesJieranShen:
@scott-degraw can we now claim explicit compatibility with a LCG release? That would be very nice, and may be worth documenting as a separate installation path (not as part of this PR obviously)

scott-degraw:
I think we can. I built it successfully with a `source /cvmfs/sft.cern.ch/lcg/views/LCG_110/x86_64-el9-gcc15-opt/setup.sh` for the dependencies. Tensorflow and nlopt are not included in the LCG though.

PR #416, opened 2026-07-22
Automatically convert procset params

azummo:
When setting a parameter using procset, ratpac decides the type from how you write the value, so 4250 is an int, and 4250.0 is a double. This can cause a parameter unknown error when the parameter is just the wrong type. For example, when using quadfitter, /rat/procset max_radius 4250 fails, but /rat/procset max_radius 4250.0 works.

This pull request changes it to automatically convert the parameter when you input an int for a double, and to throw an error when you input a double for an int.

Also, a few processors did not throw a parameter unknown error when given an unknown parameter, so that has been added. Note that this means an existing macro with an unknown parameter for these processors will now fail instead of doing nothing with the parameter.


rpitelka (approved):
Tested and looks good to me!

PR #417, opened 2026-07-24
Make AmBe neutron multiplicity configurable via ratdb

azummo:
Changes AmBeSource.cc to read in the n_neutron field from AMBE_NSPECTRUM.ratdb file instead of hardcoding 1 neutron. This allows you to set it to 0 to simulate just the 4.4 MeV AmBe gamma with the AmBe generator. (You can already simulate just the neutron by setting prob_gamma_emission to 0)


gmwendel:
A minor comment: This change would break old AMBE_NSPECTRUM default tables without the new field and generate an uncaught DBNotFoundError at the first AmBe event. 

other generators generators (Gen_LED.cc, VertexGen_Decay0.cc) use a try/catch-with-default for this situation
```
try {
Nneutron = neutrondb->GetI("n_neutron");
} catch (DBNotFoundError &e) {
Nneutron = 1;  // legacy default or a warning to the user
}
```
Perhaps this is not a big deal if everyone using the generator is aware of the change.

The actual results produced by the new code look correct
image
image

tannerbk:
@gmwendel is this good to go?

gmwendel:
LGTM

PR #418, opened 2026-07-28
Add triggerSumPeak to SplitEVDAQProc and OutNtupleProc

azummo:
This PR calculates the peak value of the trigger sum (in units of NHit) in SplitEVDAQProc and then saves it to the ntuple. This allows the user to set a low threshold and then easily apply a trigger efficiency curve in analysis afterward.

Because this value is calculated per EV, events that don't trigger the detector (or don't use SplitEVDAQProc) are set to 0. Saving some value for untriggered events would mean editing the DS::MC in a way that it currently isn't used.


JamesJieranShen (approved):

JamesJieranShen:
Thanks fro making this! I have no objection to the implementation, but I would think it is better to name this something more generic in ratpac2, such as simply triggerPeak (parallel to digitPeak), since triggerSumPeak is very MTCA-esque.

PR #419, opened 2026-07-29
Track energy and photon truth information

scott-degraw:
This PR adds the following features:

## Deposited and quenched energy for each track step 
This can be useful for seeing how much light is deposited into a certain volume. 

## Photon truth information
For each MC photoelectron the creation position/time of the original photon is stored as well as the excitation time (ie the time at which the scintillator was first excited independent of the random scintillator timing). This can be useful for creating more accurate time residual PDFs.

## Track simplification
Most of the track steps are much smaller than what is needed for downstream analysis. A processor to compact the tracks to a minimum length/time step difference is added. 

## Optional branch pruning
Tracking/MC PE information can produce a lot of branches (potentially nested branches) most of which are probably not useful for a specific analysis. The ntuple processor can be configured `/rat/procset prune "branchA,branchB*"` to filter branches. Wildcards `*` and `?` are both accepted.


scott-degraw:
Following on from @JamesJieranShen 's comment on the convenor call, I've implemented an online variant of the track compaction. This way the full track trajectory is never stored in memory. I've also adjusted it so that it always respects Geant4 geometry boundaries and doesn't merge steps across them. 

There is also now the ability to set different minimum lengths and times on a per particle basis.

Since the online variant is strictly better than the previous processor I implemented, I removed the track simplify processor.

JamesJieranShen:
@scott-degraw thank you! Would you mind resolving the conflict? Meanwhile I will go through the rest of the PR.

src/io/include/RAT/OutNtupleProc.hh
```diff
@@ -85,7 +88,24 @@ class OutNtupleProc : public Processor {
   std::unique_ptr<RAT::TransitTimeCalculator> fTransitTimeCalculator;
 
  protected:
+  bool IsPruned(const std::string &name) const {
```
JamesJieranShen: I am not sure what we should use here, POSIX match or regex. Obviously POSIX matching is a lot more intuitive since people use it everyday, but it is not portable to certain platforms, and other places in RAT such as regexfill uses regex.

src/io/include/RAT/OutNtupleProc.hh
```diff
@@ -85,7 +88,24 @@ class OutNtupleProc : public Processor {
...
+    for (const auto &pattern : prunedBranches) {
+      if (fnmatch(pattern.c_str(), name.c_str(), 0) == 0) return true;
+    }
+    return false;
+  }
+
+  template <typename T>
+  TBranch *MakeBranch(TTree *tree, const std::string &name, T *ptr) {
```
JamesJieranShen: we should allow users to still provide a leaflist if needed, for storing certain items like a c-array.

JamesJieranShen:
I've gone through the code and it looks fine -- not giving an official review since I'm not assigned to this PR. But I would suggest splitting this PR into two -- one for the `prune` changes for OutNtuples and one for the track compactor. We squash merge PRs in Ratpac-two and having both features here land in the same commit may be a bit excessive. However this should not be a blocker.

src/io/include/RAT/OutNtupleProc.hh
scott-degraw: I could see how the regex is useful but I think that should be separate functionality in a command like `prune_regex` instead of the regular `prune`. For the majority of cases `*` and `?` will do fine I think.

JamesJieranShen: Ok, I think this is fine. I was worried about  using POSIX calls, but this is apparently not the first instance where we do this, so no worries I guess..

tannerbk:
This looks good to me.

JamesJieranShen (approved):

PR #420, opened 2026-07-29
Prevent procset parameters from being overwritten by ratdb values

### LINKED ISSUE #264: Default ratdb entry prioritized over procset in macro
`/rat/procset` command in macro files seem to run successfully, but happens before the `BeginOfRun` function is called.
Hence the default ratdb entry always overwrites the procset command. Ideally procset should have priority.
This behavior has only been observed for the splitevdaq processor so far.

azummo:
Right now, some values that can be set using /rat/procset also have default values in a ratdb table. Because BeginOfRun runs after the macro is read, these default values currently overwrite the user input values.

This PR adds MarkParamSet() and WasParamSet() to track whether the /rat/procset parameter was set by the user, and prevent it from being overwritten.

I've tried to find and update every instance in ratpac-two where this applies. Quad and Mimir had their own methods for checking whether a parameter was set, and I've updated them to use MarkParamSet() and WasParamSet() while trying to keep the same behavior.

### DIFF
--- a/src/fit/src/FitQuadProc.cc
+++ b/src/fit/src/FitQuadProc.cc
@@ -15,16 +15,19 @@ void FitQuadProc::BeginOfRun(DS::Run *run) {
 
   DB *db = DB::Get();
   DBLinkPtr quad_db = db->GetLink("FIT_QUAD");
-  fNumQuadPoints = quad_db->GetI("num_points");
-  fMaxQuadPoints = quad_db->GetI("max_points");
-  fTableCutOff = quad_db->GetI("table_cut_off");
+  if (!WasParamSet("num_points")) fNumQuadPoints = quad_db->GetI("num_points");
+  if (!WasParamSet("max_points")) fMaxQuadPoints = quad_db->GetI("max_points");
+  if (!WasParamSet("table_cut_off")) fTableCutOff = quad_db->GetI("table_cut_off");
   if (fTableCutOff > fNumPointsTbl.size()) {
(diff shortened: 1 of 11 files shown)


src/fit/src/FitQuadProc.cc
```diff
@@ -15,16 +15,19 @@ void FitQuadProc::BeginOfRun(DS::Run *run) {
...
+  if (!WasParamSet("table_cut_off")) fTableCutOff = quad_db->GetI("table_cut_off");
   if (fTableCutOff > fNumPointsTbl.size()) {
     Log::Die("Quad::BeginOfRun: cannot set a table_cut_off larger than the size of fNumPointsTbl.");
   }
-  fMaxRadius = quad_db->GetD("max_radius");
-  fMaxX = quad_db->GetD("max_x");
-  fMaxY = quad_db->GetD("max_y");
-  fMaxZ = quad_db->GetD("max_z");
+  // Only use defaults if none were set
+  if (!(WasParamSet("max_radius") || WasParamSet("max_x") || WasParamSet("max_y") || WasParamSet("max_z"))) {
```
llebanowski, line 25: max_radius should be tested separately from max_x/y/z because it can be set independently.  thanks.

llebanowski: looking more carefully, this looks good as is.  thanks again.

syjeon162:
This looks good to be merged once the merge conflicts are resolved.

PR #421, opened 2026-08-01
quad: add fixed position option

llebanowski:
Add the option to return a fitted time from quad by providing a fixed position.  
The test below shows that the addition works as expected.  

Using the nominal quad macro:
Screenshot 2026-08-01 11 21 14

Using the quad macro with the fixed position (0,0,0):
Screenshot 2026-08-01 11 24 51


llebanowski:
Closing in favor PR #423

PR #423, opened 2026-08-08
Add simple time fitter

llebanowski:
This PR supercedes #421 by creating a new processor.  
Notably, it uses all available hit times and offers the user a choice between two straight-line light path options:
1. Use `light_speed` (default)
2. Use the `TransitTimeCalculator` by specifying the photon wavelength to use.  

Default:
image

TransitTimeCalculator:
image


src/fit/src/FitTimeProc.cc
```diff
@@ -0,0 +1,143 @@
...
+#include <RAT/Processor.hh>
+#include <cmath>
+#include <string>
+#include <vector>
+
+namespace RAT {
+
+void FitTimeProc::BeginOfRun(DS::Run *run) {
+  DBLinkPtr table = DB::Get()->GetLink("FIT_COMMON", "");
+  fLightSpeed = table->GetD("light_speed");
```
azummo: This should be updated to:

if (!WasParamSet("light_speed")) fLightSpeed = table->GetD("light_speed");

so that it doesn't overwrite a value set by procset

src/fit/src/FitTimeProc.cc
```diff
@@ -0,0 +1,143 @@
...
+}
+
+Processor::Result FitTimeProc::Event(DS::Root *ds, DS::EV *ev) {
+  inputHandler.RegisterEvent(ev);
+
+  DS::FitResult *fitT = new DS::FitResult(name, fFitLabel);
+  fitT->SetEnableTime(true);
+
+  /// Initialize FitResult parameters
+  fitT->SetPosition(fPosition);
```
azummo, line 80: SetPosition() calls SetValidPosition(true), so validposition_fittime is always true, even if the fit fails. I guess the position is the value input by the user, so it is arguably valid. Just wanted to check if this is the desired behavior.

src/fit/src/FitTimeProc.cc
```diff
@@ -0,0 +1,143 @@
...
+    for (double time : inputHandler.GetTimes(pmtid)) {
+      // Select PMTs by time - optional
+      if ((fSetMaxHitTime && time > fMaxHitTime) || (fSetMinHitTime && time < fMinHitTime)) {
+        continue;
+      }
+
+      const TVector3 pmtPos = fPMTInfo->GetPosition(pmtid);
+
+      double tt = 0;
+      if (fSetWavelength) {
```
azummo, line 116: The user can set both fWavelength and fLightSpeed, but if they do, only fWavelength is used. I think it should either only allow the user to set one, or it should be stated in the documentation that if both are set, the wavelength and the TransitTimeCalculator are used, and the light speed is ignored.

macros/examples/fittime.mac
```diff
@@ -0,0 +1,31 @@
+# Example macro showing how to use quadfitter
```
azummo: Typo, quadfitter->fittime

doc/users_guide/reconstruction.rst
```diff
@@ -16,6 +16,50 @@ Document the fitter handler.
...
+
+``min_hit_time``            ``double``                  Optional lower cut on PMT hit times in ns
+``max_hit_time``            ``double``                  Optional upper cut on PMT hit times in ns
+
+``event_position_x``        ``double``                  Position of event in mm.  Defaults to 0.
+``event_position_y``        ``double``                  Position of event in mm.  Defaults to 0.
+``event_position_z``        ``double``                  Position of event in mm.  Defaults to 0.
+=========================   ==========================  ===================
+
+Position fit information in data structure
```
azummo: Position fit->time fit

doc/users_guide/reconstruction.rst
```diff
@@ -16,6 +16,50 @@ Document the fitter handler.
...
+
+Parameters
+''''''''''
+All of the parameters below can be set in macro, which allows the processor to be run multiple times
+with different settings in a single macro.
+
+=========================   ==========================  ===================
+**Field**                   **Type**                    **Description**
+=========================   ==========================  ===================
+``label``                   ``string``                  Additional string appended to "fitcentroid"
```
azummo: Typo: fitcentroid->fittime

src/fit/src/FitTimeProc.cc
llebanowski: Yes, this is exactly the intended logic.

azummo:
All my comments have been addressed. Looks good to me.

PR #424, opened 2026-08-08
Standardize usage of fit->Set functions

llebanowski:
- Remove unnecessary initializations that simply overwrite those of FitResult.  
- Ensure that SetValid is called after the quantity is Set because the latter automatically sets the validity to true.  
- Ensure that relevant SetEnable functions are called immediately after the FitResult is created.

### DIFF
--- a/src/fit/src/FitDirectionCenterProc.cc
+++ b/src/fit/src/FitDirectionCenterProc.cc
@@ -94,15 +94,9 @@ Processor::Result FitDirectionCenterProc::Event(DS::Root *ds, DS::EV *ev) {
   inputHandler.RegisterEvent(ev);
 
   DS::FitResult *fitDC = new DS::FitResult(name, fFitLabel);
+  fitDC->SetEnableDirection(true);
 
-  /// Initialize ALL parameters with placeholder values
-  fitDC->SetEnablePosition(false);
-  fitDC->SetValidPosition(false);
-  fitDC->SetPosition(TVector3(0, 0, 0));
-  fitDC->SetEnableDirection(false);
-  fitDC->SetValidDirection(false);
-  fitDC->SetDirection(TVector3(0, 0, 0));
-  // Figures of Merit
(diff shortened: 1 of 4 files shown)


src/fit/src/FitDirectionCenterProc.cc
```diff
@@ -94,15 +94,9 @@ Processor::Result FitDirectionCenterProc::Event(DS::Root *ds, DS::EV *ev) {
   inputHandler.RegisterEvent(ev);
 
   DS::FitResult *fitDC = new DS::FitResult(name, fFitLabel);
+  fitDC->SetEnableDirection(true);
```
azummo, line 97: Should SetEnablePosition be done here? What happens if it should be true and the fitter returns Processor::FAIL (e.g. at line 114) before reaching SetValidPosition() at line 193.

azummo:
Looks good, just the minor comment. Also, should the documentation be updated?

src/fit/src/FitDirectionCenterProc.cc
llebanowski: I view GetEnable as telling a user whether an algorithm has been initiated and then GetValid tells a user if the result is valid.  So, yes, I think SetEnable should be set to true at the beginning here.  
If a fitter returns before setting any part of a FitResult, then a user would get default values, which includes GetValid == false.

llebanowski:
> Looks good, just the minor comment. Also, should the documentation be updated?

I think things are in order.  
I am willing to help with documentation.  Perhaps the convenors would like to start a section in doc/users_guide
/reconstruction.rst

src/fit/src/FitDirectionCenterProc.cc
llebanowski: Sorry, I misread SetEnablePosition in the comment.  
Since this fitter provides a position when applying a drive correction, I will add SetEnablePosition here along with the drive conditional.  Thanks

JamesJieranShen:
@azummo look ok?

PR #427, opened 2026-08-13
Uncomment PMT glass absorptions

llebanowski:
This PR is a follow-on of #376 , where the absorption of PMT glass models was commented out.  

Since QE measurements generally include absorption effects, excluding absorption in PMT glass models is appropriate.  However, this can confuse the use of these materials for purposes where absorption is desired.  To avoid this, absorption is being uncommented and notes placed in the ratpac-two code to more clearly inform users of how PMT glass absorptions and PMT quantum efficiencies are implemented.  Note that GLG4PMTOpticalModel does not use absorption.  
- Uncommented ABSLENGTH for quartz, hamamatsu_borosilicate_glass, & borosilicate_glass.  
- Slightly modified comment in OPTICS.ratdb and added a new comment to OPTICS_Photocathode.ratdb.  
- Moved a few definitions (mostly PMT glasses) for more logical grouping.  Sorry for the messy diff.  
- (Removed `dead nobles' which have no optical definition and should not be implemented in that way anyway.)


tannerbk:
As you point out, the diff is confusing because of the rearranged materials. This looks fine to me.

PR #428, opened 2026-08-18
Add guard to GetTime

rpitelka:
When a waveform analyzer does not reconstruct any PEs for a given DigitPMT, any event fitters that call `GetTime` from the `FitterInputHandler` will cause a crash because `GetTime` expects a non-empty vector of times. 

To fix this, I added a check to `GetTime` in `FitterInputHandler` that dies when the function is called when no PEs were reconstructed. I also added checks to all the fitters that call `GetTime` to skip PMTs with no reconstructed PEs.

I'm not sure this is the best solution, so input is appreciated, but it does fix the known crash when empty waveform analysis results from RAVEN (caused by the `max_total_charge` parameter being set too low) are fed into quad.

### DIFF
--- a/src/fit/include/RAT/FitterInputHandler.hh
+++ b/src/fit/include/RAT/FitterInputHandler.hh
@@ -313,7 +313,11 @@ class FitterInputHandler {
         if (std::find(fitterNames.begin(), fitterNames.end(), wfm_ana_name) == fitterNames.end()) {
           info << "FitResult not found for pmt id " << id << " " << wfm_ana_name << newline;
         }
-        return digitpmt->GetOrCreateWaveformAnalysisResult(wfm_ana_name)->getTime(0);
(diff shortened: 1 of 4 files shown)


JamesJieranShen:
Discussed offline -- I'm not sure if throwing out all the hits above max charge is the best option. I think it may be better to have InputHandler fall back to `digit*` when the fit returns empty, needs more thinking....

JamesJieranShen:
Thanks @rpitelka , are the changes to FitDirectionCenter now obsolete?

rpitelka:
@JamesJieranShen  Fixed revert, now only `FitterInputHandler` is modified by this PR

JamesJieranShen (approved):

syjeon162:
Looks good! Can we add some documentation about this behavior wherever relevant? (Maybe the `FIT_COMMON` or `DIGITIZER_ANALYSIS` ratdbs?)

syjeon162 (approved):

PR #430, opened 2026-08-25
Add pileup generator

scott-degraw:
I've added a new generator to simulate various forms of pileup. It uses the standard vertex, position and time generators with `/generator/add pileup VERTEX:POSITION:TIME` but also allows you generate different multiplicities with `/generator/pileup/multiplicity` and different timings of the pileup events with `/generator/pileup/timing`. 

The multiplicity has three distribution options: fixed, uniform, poisson. The timing has three independent distributions of fixed (constant spacing between vertices), uniform (each time sampled in a uniform distribution) and poisson (exponential time distribution).

### DIFF
--- a/doc/users_guide/generators/toplevel/pileup.rst
+++ b/doc/users_guide/generators/toplevel/pileup.rst
@@ -33,9 +33,9 @@ poisson mean must be greater than 0.
 
 ::
 
-    /generator/pileup/spacing fixed:dt
-    /generator/pileup/spacing uniform:min:max
-    /generator/pileup/spacing poisson:mean
+    /generator/pileup/timing fixed:dt
+    /generator/pileup/timing uniform:min:max
+    /generator/pileup/timing poisson:mean
 
 Sets the distribution, in ns, for vertex times relative to the event start
 time (TIME). ``fixed`` places vertices dt apart, starting at TIME.
@@ -51,7 +51,7 @@ Example
 ::
 
     /generator/pileup/multiplicity poisson:5
-    /generator/pileup/spacing uniform:0:1000
+    /generator/pileup/timing uniform:0:1000
 
 A constant-rate radioactive source observed over a fixed time window: the
 number of decays in the window is Poisson-distributed, and conditioned on
--- a/doc/users_guide/generators/toplevel/pileup.rst
+++ b/doc/users_guide/generators/toplevel/pileup.rst
@@ -1,0 +1,59 @@
+pileup
+''''''
+::
+
+    /generator/add pileup VERTEX:POSITION:TIME
+
+or
+
+::
+
+    /generator/add pileup VERTEX:POSITION
+
+Creates a new pileup generator using the vertex and position generators
+described below. If the variant without a TIME parameter is used, it implies
+the "poisson" time generator. The TIME generator controls the spacing
+between separate pileup *events*; within each event, the generator produces
+multiple primary vertices from the same VERTEX/POSITION configuration,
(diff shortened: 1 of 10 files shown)


doc/users_guide/generators/toplevel/pileup.rst
```diff
@@ -0,0 +1,59 @@
...
+Sets the distribution, in ns, for vertex times relative to the event start
+time (TIME). ``fixed`` places vertices dt apart, starting at TIME.
+``uniform`` draws each vertex's offset from TIME independently and
+uniformly between min and max, then time-orders the results.
+``poisson`` draws each vertex's offset from TIME independently from an
+exponential distribution with the given mean, then time-orders the results.
+Defaults to ``fixed:0``.
+
+Example
+--------
```
azummo, line 49: I think this line creates a new section and breaks the following sections in the documentation. Just removing it should fix things.

macros/examples/pileup.mac
```diff
@@ -0,0 +1,34 @@
...
+#
+
+/glg4debug/glg4param omit_muon_processes 1.0
+/glg4debug/glg4param omit_hadronic_processes 1.0
+
+/rat/db/set DETECTOR experiment "Validation"
+/rat/db/set DETECTOR geo_file "Validation/Valid.geo"
+
+/run/initialize
+
```
azummo, line 16: All of the other example macros include a trigger here. Usually splitevdaq.

src/gen/src/PileupGen.cc
```diff
@@ -0,0 +1,254 @@
...
+}
+
+std::vector<double> PileupDist::GenerateTimes(uint64_t n, double startTime) const {
+  std::vector<double> times(n);
+  switch (type) {
+    case kPileupFixed:
+      for (uint64_t i = 0; i < n; i++) times[i] = startTime + i * fixedValue;
+      break;
+    case kPileupUniform:
+      for (uint64_t i = 0; i < n; i++) times[i] = startTime + CLHEP::RandFlat::shoot(uniformMin, uniformMax);
```
azummo, line 87: When simulating uniform pileup, I assume that you would want an equal chance of the time difference of the two events to be any value in [min,max], but drawing two random values from [min,max] does not give you that.

src/gen/src/PileupGen.cc
```diff
@@ -0,0 +1,254 @@
...
+      std::sort(times.begin(), times.end());
+      break;
+    case kPileupPoisson:
+      for (uint64_t i = 0; i < n; i++) times[i] = startTime + CLHEP::RandExponential::shoot(poissonMean);
```
azummo, line 91: For Poisson, the timing between the events should be exponential, but this pulls multiple times from a single exponential distribution. The documentation describes what the code does, but this should probably either be renamed "exponential" or changed to actually be Poisson distributed.

azummo:
I've run the generator, and it seems to work and simulate pileup events.

In addition to the comments I've added, as a general question: How different is this from the existing Coincidence generator? Would it make more sense to add these features to that generator rather than create a new one?

scott-degraw:
Thanks for looking at this Anthony. The main difference is the addition of randomised multiplicities. I suppose I could add this to the coincidence generator. However, I think the current coincidence generator code is overly complex so I would probably just end up rewriting all of the core logic.

scott-degraw:
I think this generator might have too many useless options available. The main idea was to use the poisson multiplicity generator with the uniform timing generator to simulate a poisson process over a fixed time window. This should be just simplified to a fixed interval, poisson and exponential options.

PR #431, opened 2026-08-25
Enforce absolute path requirement for rattest experiment binaries

rpitelka:
The experiment binary that can be provided for a rattest should be an absolute path, but this is not explicitly enforced, causing inconsistent errors downstream if a binary name is provided instead. Added an explicit check and error in the case that an absolute path is not provided.


rpitelka:
Offline discussion suggested allowing the pre-existing behavior to function properly by calling `shutil.which` on `rat_dir` instead of `os.path.abspath`.

PR #433, opened 2026-08-26
Add extra paths for ratdb

scott-degraw:
RATDB currently only looks for `.ratdb` table files in `$RATSHARE/ratdb`. This adds a new environment variable, `RATDB_EXTRA_PATH`, that lets users point RAT at additional directories.
- `RATDB_EXTRA_PATH` is a colon-separated list of one or more directories (like `$PATH`).
- Directories listed take priority over `$RATSHARE/ratdb`: a table with the same name/index in one of these directories overrides the corresponding default. Among multiple directories, earlier-listed ones take priority over later ones.
- If a listed directory doesn't exist (or isn't a directory), RAT prints a warning and skips it rather than failing silently or hard-erroring.
- Behavior is unchanged when `RATDB_EXTRA_PATH` is unset.
- Previously the ratdb directories were stored in a `std::set` so the ratdb priority was lexicographic. Directories are now stored in a `std::vector` so that any order can be enforced

### DIFF
--- a/src/ratbase/include/RAT/Rat.hh
+++ b/src/ratbase/include/RAT/Rat.hh
@@ -5,6 +5,7 @@
 #include <RAT/AnyParse.hh>
 #include <RAT/DB.hh>
 #include <RAT/DBMessenger.hh>
+#include <RAT/Log.hh>
 #include <RAT/ProducerBlock.hh>
 #include <RAT/RatMessenger.hh>
 #include <algorithm>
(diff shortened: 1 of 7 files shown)


scott-degraw:
The change of `ratdb_directories` to an `std::vector` caused downstream experiments that accessed that variable to fail. To counteract this I made a new `RatdbDirectoryList` class to keep an ordered list of ratdb directories in `ratdb_search_path` but I still keep the old `ratdb_directories` as a `std::set` so downstream code does not break. `ratdb_directories` and `ratdb_search_path` are always kept in sync but only the new `ratdb_search_path` has a predictable priority ordering.

src/ratbase/include/RAT/Rat.hh
```diff
@@ -7,11 +7,52 @@
...
+ *  directory in reverse order and letting later loads overwrite earlier ones
+ *  makes the highest-priority directory win (see DB::LoadDefaults()). Every
+ *  begin()/rbegin() call re-pulls ratdb_directories' current contents into
+ *  this combined view; don't hold an iterator across a fresh begin()/rbegin()
+ *  call, since it rewrites the storage that iterator points into. Normal
+ *  iteration (which calls begin()/rbegin() once, then end()/rend() once) is
+ *  unaffected.
+ **/
+
+class RatdbDirectoryList {
```
JamesJieranShen, line 37: While performance is of little concern here, it may be more idiomatic to use `std::list` as the internal container rather than a vector. Additionally it should make the prepend/append logic simpler to write.

src/ratbase/include/RAT/Rat.hh
```diff
@@ -7,11 +7,52 @@
...
+
+  std::vector<std::string>::const_iterator begin() const;
+  std::vector<std::string>::const_iterator end() const { return combined_.end(); }
+  std::vector<std::string>::const_reverse_iterator rbegin() const;
+  std::vector<std::string>::const_reverse_iterator rend() const { return combined_.rend(); }
+
+ private:
+  void RebuildCombined() const;
+
+  std::vector<std::string> dirs_;
```
JamesJieranShen: we should follow G4 convention/rest of RAT code to use `fDirs` etc. for member variables.

src/ratbase/src/Rat.cc
```diff
@@ -105,6 +103,29 @@ void Rat::Configure() {
...
+  // RATDB_EXTRA_PATH is a colon-separated list of directories searched (and
+  // loaded) with priority over $RATSHARE/ratdb (appended below, so it's
+  // always lowest priority), highest priority first.
+  if (getenv("RATDB_EXTRA_PATH") != NULL) {
+    std::stringstream extra_path(static_cast<std::string>(getenv("RATDB_EXTRA_PATH")));
+    std::string dir;
+    while (std::getline(extra_path, dir, ':')) {
+      if (dir.empty()) continue;
+      struct stat s;
+      if (stat(dir.c_str(), &s) != 0 || !S_ISDIR(s.st_mode)) {
```
JamesJieranShen: let's use stl: `std::filesystem::is_directory`

JamesJieranShen (commented):
Thanks @scott-degraw , this has always been a pain point in RATPAC2 and I'm happy to see that this is getting better! Please see line comments.

Additionally, for your implementation of backward compatibility, I'm wondering if it is worthwhile to simply expose the legacy variables of `ratdb_directories` etc. simply as aliases to the new struct, and implement an `insert` interface that effectively resolves to Append/Prepend. The lexicographical sorting of `std::set` was always an unfortunate side effect and we never guaranteed that override behavior (in fact, we specifically state in the doc that this sort of override is unsupported), so preserving that is not entirely necessary in my opinion. Downstream can still insert overrides via experiment-specific directories. 

A longer discussion should be had in terms of how we would like to handle RATDB overrides again, now that we have a better way to control the ordering with this PR.

Thanks again!

src/ratbase/include/RAT/Rat.hh
scott-degraw: I think an `std::vector` is actual quicker here just because this list should never be more than a few elements and an `std::vector` is lighter weight and contiguous compared to a double linked list. The prepend/append logic is the same either way; `std::vector` can be replaced with `std::list` with no other code changes.

JamesJieranShen: I agree on your comment on performance. I was more thinking that the `std::erase(std::remove())` syntax may be avoidable in a linked list so the code becomes a bit more readable. But this is a very minor issue, feel free to resolve.

scott-degraw:
I think I could keep `ratdb_directories` and just add an `insert` method to my new class. Strictly speaking this doesn't preserve backwards compatibility because downstream code could assume other `std::set` properties but maybe that's being too cautious.

JamesJieranShen:
> I think I could keep `ratdb_directories` and just add an `insert` method to my new class. Strictly speaking this doesn't preserve backwards compatibility because downstream code could assume other `std::set` properties but maybe that's being too cautious.

I do think it is likley overly cautious: there's a not a lot of downstream experiments, certainly not ones that are extremely customized. Having ratdb_directories be a reference to your new class and allow insert to be an append/prepend eliminates a lot of complexity that was inserted for the sake of legacy support, which I think we should try to avoid.

scott-degraw:
I've removed `ratdb_search_paths` and added a legacy `insert` property which gives a warning and calls `Prepend` which I think is the right default choice for downstream code. The logic is simpler now.

JamesJieranShen (approved):

PR #437, opened 2026-09-08
Explicitly include `<limits>`

### LINKED ISSUE #436: Implicit `<limits>` include failing with newer GCC version
Using up-to-date GCC versions runs into a compilation error in `src/util/src/WaveformUtil.cc`:

```
/build/source/src/util/src/WaveformUtil.cc: In function 'double RAT::WaveformUtil::CalculatePedestalSlidingWindowADC(const std::vector<short unsigned int>&, int)':
/build/source/src/util/src/WaveformUtil.cc:259:24: error: 'numeric_limits' is not a member of 'std'
  259 |   double minVar = std::numeric_limits<double>::max();
      |                        ^~~~~~~~~~~~~~
/build/source/src/util/src/WaveformUtil.cc:259:39: error: expected primary-expression before 'double'
  259 |   double minVar = std::numeric_limits<double>::max();
```

This is due to `std::numeric_limits` requiring the `#include <limits>` import. In older GCC versions, this import is done implicitly. However, newer versions of GCC have (correctly) eliminated this obfuscated import, requiring `<limits>` to be included explicitly.

This fix should be as easy as adding `#include <limits>` to `WaveformUtil.cc`. I haven't been able to pinpoint the exact version of GCC where this breaks.

mdcge:
This will work with modern GCC libstdc++ versions, which have fixed an transitive inclusion problem with `limits`.

Fixes #436


mdcge:
Tried a couple things, but haven't managed to get the format check to pass.

How many ways can you incorrectly add a single line of code... :(

JamesJieranShen:
Fixed. For future reference, you can just install clang-format on your local machine and run `./cformat.sh`.  If this is installed, it should also automatically do this before every commit!

JamesJieranShen (approved):

PR #439, opened 2026-09-08
Skip `RAT::Configure` on repeat call

### LINKED ISSUE #438: Custom processors registered before RATDB loaded in `RAT::Configure`
Experiment-specific `rat` executable fails with `RAT::DBNotFoundError` thrown from `OutNtupleProc`'s constructor.

I have chased this down to the following: experiment-specific processors are loaded *before* the general `RAT::Configure()` runs (which is called from `RAT::Begin()`). `Configure` is what loads RATDB tables. Therefore, if an experiment-specific processor attempts to access RATDB (which is typically the case for any custom versions of `OutNtupleProc`) this will fail as RATDB is not initialised yet.

I can imagine two solutions:
1. Place the RATDB access lines (such as `event_fitters = table->GetSArray("event_fitters");`) inside the `try`-`catch` blocks that follow.
2. Allow `Configure` to be called by the experiment-specific constructor (i.e. allow `Configure` to be called multiple times), before the custom processors are registered.

mdcge:
This allows `Configure` to be invoked before `RAT::Begin`. This is useful for custom processors which require RATDB to be loaded before `RAT::Begin` is called.

Fixes #438


JamesJieranShen:
Hi @mdcge , thanks for this. I'm a bit confused about this issue in general. Are you trying to call RATDB in the constructor of the processor? The canonical resolution to this issue is to simply configure the processor in `BeginOfRun` rather than the constructor.

PR #443, opened 2026-09-13
CMake refactor

scott-degraw:
# CMake refactor

This is a fairly significant refactor of the CMake build system. 

## Target based

The main improvement is using modern target based usage requirements instead of the hardcoded "make style" commands from before. Most of the core dependencies are now packaged into the `ratpac_common` target. Each module's `CMakeLists.txt` goes from

    target_include_directories(ratbase SYSTEM PUBLIC 
            $<BUILD_INTERFACE:${RATPAC_INCLUDE_DIR}/stlplus>)
    target_include_directories(ratbase PUBLIC
            $<BUILD_INTERFACE:${RATPAC_INCLUDE_DIR}>
            $<INSTALL_INTERFACE:include>)

to

    target_link_libraries(ratbase PUBLIC ratpac_common)

which includes all include, linking, preprocessor definitions (not properly implemented before) and more to the target. Optional module dependencies can now be more easily included e.g.

    target_link_libraries(mimir PUBLIC ratpac_common Ratpac::NLopt)

The new `cmake/RatpacDependencies.cmake` finds all the dependencies to define the new `ratpac_common` and `Ratpac::...` targets.

## Configure time copies

The configure time copy of all the module include files to a single `include` directory has been removed and each module objects library gets the include directories through `ratpac_common`. This removes a bug where modifying a header file and building without reconfiguring would not include the header file changes. The install target still copies all the headers to a single `include` when installing but this is done using `install` functions instead of the configure time copy. The `install` command now also means that the downstream `Ratpac` target is a proper target with target dependencies instead of hardcoded strings.

The generated config header files are now configured straight into the binary directory. This means the awkward `.gitignore` entries (e.g. `src/core/config.hh`) are removed. `ratpac_common` has the generated header files properly included.

## libRATEvent

I'm not sure if this shared library was ever used very much (is it just genie?) but it was causing build issues for me on my mac. The issue was that it was never linked properly and had undefined symbols due to incomplete linking declarations. You can see this by running `ldd -r libRATEvent.so`. Linux's linker ignored this waiting until runtime whereas macOS's linker refused to accept this at link time. The workaround was to just link it with the same targets as RATPAC in which it just becomes byte-by-byte identical to libRATPAC. If this is meant to be a lightweight version of the RATPAC library, some source changes will need to be made so that it can be linked against a subset of ratpac.

## CMake version bump

Min version bump from 3.11 to 3.16


CMakeLists.txt
```diff
@@ -58,117 +58,56 @@ add_custom_target(
 ###########################################################
 # Set default build type
 set(default_build_type "RelWithDebInfo")
-#set(CMAKE_BUILD_TYPE Debug)
```
JamesJieranShen, line 61: Can we keep these lines as comments? I think many folks may be accustomed to un-commenting these lines and continue using the Makefile to switch build modes

CMakeLists.txt
```diff
@@ -58,117 +58,56 @@ add_custom_target(
...
+    "Debug" "Release" "MinSizeRel" "RelWithDebInfo"
+  )
 endif()
 add_compile_options(-Wall)
 # Not turning on errors yet to ease transition
 # add_compile_options(-Werror)
 
 ###########################################################
 # Set CMAKE standards
-set(CMAKE_CXX_STANDARD 17)
```
JamesJieranShen: I think we do want these, since rp2 explicitly uses cxx17 features

CMakeLists.txt
```diff
@@ -186,49 +125,64 @@ add_subdirectory(tools)
...
+  ${PROJECT_BINARY_DIR}/bin/ratpac.csh
+  @ONLY NEWLINE_STYLE UNIX
+)
+# The package is installed twice, at two different destinations, on purpose:
+# DESTINATION cmake (flat) is this project's original layout, and is what
+# existing consumers (e.g. RatpacExperiment) already point at; DESTINATION
+# lib/cmake/Ratpac is the conventional CMake package layout, which lets a
+# plain, unhinted find_package(Ratpac) also succeed via CMake's default
+# package search paths. RatpacConfig.cmake has to be configured once per
+# destination
```
JamesJieranShen: Can we rewrite this comment a bit..? The LLM is coming a bit strong here :)

src/CMakeLists.txt
```diff
@@ -1,116 +1,116 @@
...
-add_subdirectory(gen)
-add_subdirectory(geo)
-add_subdirectory(io)
-add_subdirectory(physics)
-add_subdirectory(external)
-add_subdirectory(util)
-add_subdirectory(ds)
-add_subdirectory(ratbase)
-
+###########################################################
```
JamesJieranShen: The logic of this hunk really should be written as part of the boiler plate for each directory, if possible. This way, custom logic (such as MIMIR's inclusion) can just be extra lines contained within one directory, not being specified in the global CMake config.

Not sure if you've seen https://github.com/snoplus/rat/blob/9.0.0alpha1/src/CMakeLists.txt, but setting up a similar macro to `add_rat_header` to abstract away some boilerplate code may be useful here as well.

src/CMakeLists.txt
```diff
@@ -1,116 +1,116 @@
...
+# Get every module's include directory in {module_name}/include if it exists
+set(RATPAC_MODULE_INCLUDE_DIRS)
+foreach(_dir IN LISTS RATPAC_MODULE_DIRS)
+  if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${_dir}/include)
+    list(APPEND RATPAC_MODULE_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/${_dir}/include)
+  endif()
+endforeach()
+# Add on the nested mimir include and the generated include 
+list(APPEND RATPAC_MODULE_INCLUDE_DIRS
+    ${CMAKE_CURRENT_SOURCE_DIR}/fit/mimir/include
```
JamesJieranShen: For this specficially -- MIMIR is really designed to only expose its (one) RAT header. If it makes things easier, feel free to move `RAT/FitMimirProc` to the common directory, and keep mimir's other headers private.

src/CMakeLists.txt
```diff
@@ -1,116 +1,116 @@
...
-       io/src/DSReader.cc
-       io/src/DSWriter.cc
-       io/src/TrackCursor.cc
-       io/src/TrackNav.cc
-       util/src/ReadFile.cc)
-target_include_directories(RATEvent SYSTEM PUBLIC ${ROOT_INCLUDE_DIRS}/stlplus)
-target_include_directories(RATEvent PUBLIC ${RATPAC_INCLUDE_DIR})
-target_link_libraries(RATEvent PUBLIC 
-       ${Geant4_LIBRARIES} ${ROOT_LIBRARIES} RATDict stlplus)
+### library for ROOT
```
JamesJieranShen, line 65: RATEvent was recently deprecated by me in favor of linking the entire libRATPAC with ROOT. I'm just now realizing that this was actually always broken, but kind of working (until someone called a symbol they are not supposed to). Personally, I'm in favor of completely removing it -- nothing should need to use it anymore, and this double-build is completely unnecessary. but we should come to a decision in a RAT call.

cmake/RatpacDependencies.cmake
```diff
@@ -0,0 +1,137 @@
...
+  target_link_directories(Ratpac::CRY INTERFACE ${CRYLIBDIR})
+endif()
+
+###########################################################
+# ratpac_common
+#
+# Baseline usage requirements every RAT module needs: C++17, the collected
+# header tree, Geant4, and ROOT.
+#
+add_library(ratpac_common INTERFACE)
```
JamesJieranShen: This target (along with the CMake standards etc) really should be defined in the global CMakeList. I think it's fine to declare ROOT/G4 as dependencies in this file, though.

Also, We should use all caps to be consistent with the rest of RATPAC's global targets. RATPAC_INCLUDES may be a better name than ratpac_common, which IMO is not very descriptive. But that may be more so for taste.

src/CMakeLists.txt
```diff
@@ -1,116 +1,116 @@
...
+set(RATPAC_MODULES
+  cmd core daq db ds gen geo io fit physics ratbase stlplus cppflow util RATDict
+)
 
+# Create the shared RATPAC target for downstream experiments
+# Add third party dependencies as PUBLIC and include all of the ratpac 
+# code as PRIVATE  
+add_library(RATPAC SHARED)
+target_link_libraries(RATPAC
+  PUBLIC Ratpac::ROOT Ratpac::Geant4 Ratpac::CRY Ratpac::TensorFlow Ratpac::NLopt
```
JamesJieranShen, line 58: This is fine, but I think there's a potential issue now with the ordering of how includes are made in Eos/RatpacExperiment. It would be great if these could be corrected to conform to the new dependency setup!

JamesJieranShen (commented):
Thanks @scott-degraw , for your hard work as always! I did this myself when I was working on SNO+ RAT's cmake build chain, and obviously have been hit by the bad compile_command reference resolution a couple times myself, so it is good to see this fixed in ratpac2. Some comments on the specific places certain lines are put, mostly to make things a bit more maintainable. As mentioned in the review, it would be good to see some relevant changes being propagated to the experiment template and Eos.

I've tested this myself against the latest version of EosSimulations, and all works well.

CMakeLists.txt
scott-degraw: Generally I think we shouldn't have commented lines in production code. The makefile already has a `make debug` mode. I can put a comment in that position telling the user about that?

src/CMakeLists.txt
scott-degraw: That makes sense, I've added that.

src/CMakeLists.txt
scott-degraw: That makes things a little easier. The public header called one of the internal headers so I did have to do a forward declaration.

cmake/RatpacDependencies.cmake
scott-degraw: I think the lower case makes sense here because it is an convenience `INTERFACE` target only used to accumulate different dependencies. If you try to build it with cmake nothing will happen unlike the other targets that have capitalisation. I think `RATPAC_INCLUDES` might also be mistaken for the `RATPAC_INCLUDES_DIR` variable that I removed.

src/CMakeLists.txt
scott-degraw: What's the issue with the ordering? I'm not sure I understand.

JamesJieranShen: Eos, for example, attempts to find_package(GEANT4) first before attempting to find ratpac2. I think this can potentially get us into a situation where the experiment is linking to a different version of GEANT as ratpac2, so likely one should be looking for RAT first?

